PR #5, opened 2019-10-03
Python API

### LINKED ISSUE #2: Spin in Python API
Currently, there is no way to specify the spin (`--uhf` flag) using the Python API. Are there any plans to add this feature? I can look through the source code now that it's publicly available and see if I can get it to work if not.

awvwgk:
This will bring the `xtb` Python API on the same level as the `dftd4` Python API. Suggestion on what to support are welcome.

- [x] create a Python module + `setup.py` for `xtb`
- [x] add tests for `pytest` and enforce it in `meson.build`
- [x] add type hinting as necessary using `mypy` and enforce it in `meson.build`
  - [x] Python 3.5 compatibility
- [x] find an alternative to `ctypes.util.find_library` and `numpy.ctypeslib`
- [x] enforce successful `pylint` and `flake8` checks in `meson.build`
- [x] allow multiplicity in API (fix #2)
  - [x] add sanity check for number of electrons and multiplicity (either fix ~or error on wrong input~)
- [x] separate optimization script from Python module
- [x] make use of optional arguments on both sides of the API
- [x] merge PEEQ and GFN0 into one interface (same backend is used anyway)
- [x] add GBSA solvation for GFN0 via API
- [x] allow GFN1 interface to return partial charges, dipole moments, bond orders, ...
- [x] add a wrapper for GBSA
  - [x] make GBSA properties (like Born-radii and SASA) available in Python
  - [x] make GBSA solvation free energy calculations possible

This pull request is breaking API compatibility.


awvwgk:
Seems to be ready for review/merge now.

PR #15, opened 2019-10-07
add contribution guidelines to xtb

awvwgk:
Since we already took the step to go open source, we should also commit to the idea of an open source community. I wrote up contributing guidelines for `xtb`.

One open question: should we enforce a Developer Certificate of Origin (DCO, see https://probot.github.io/apps/dco/ for an easy way to implement it) with `xtb`? It seems pretty standard for larger projects and is more developer friendly than requiring to sign a Contributor License Agreement (CLA).

### DIFF
--- a/.github/dco.yml
+++ b/.github/dco.yml
@@ -1,0 +1,2 @@
+require:
+  members: false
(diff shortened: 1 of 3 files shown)


awvwgk:
Just added everybody with an account as co-author of this commit such that we can correctly populate our contributors list at GH.

PR #38, opened 2019-11-01
remove GFN parameters from global storage

awvwgk:
This tries to remove the `aoparam` module from `xtb`. Currently all parameters are stored in this global module which makes working with the parametrisation kind of annoying. Also after introducing #10 (2b4a896) the build time of `xtb/aoparam.f90` increased significantly.

- [ ] new Hamiltonian class holding all parameters
  - [ ] global parameters from `scc_parameters` object
  - [ ] element-wise parameters from `aoparam`
  - [ ] element-pairwise parameters from `aoparam`
  - [ ] shell-wise parameters from `aoparam`
- [x] real parameters are moved from `aoparam` to `mctc_param`
  - [x] better parameter storage in `mctc_param`


aslozada:
Hi!
The parameters, *.inc files (include directory), will contained in mctc_param module?

awvwgk:
Some former `aoparam` arrays moved to `mctc_param` since they should have been there from the beginning. For the fade of the `include` files, I guess nobody is going to voluntarily touch them again, so they will be internally copied to the `tb_hamiltonian` object and than left alone.

awvwgk:
Closing this, since I have no time to continue working on it.

Issue #41, opened 2019-11-05
Add the number of point charge as the first line of the pcgrad


xiki-tempula:
**Is your feature request related to a problem? Please describe.**
When using the QM/MM Orca interface of amber, it is required that the first line of the pcgrad should indicate the number of point charge in the system, which is not present in the pcgrad file.

**Describe the solution you'd like**
Normal Orca pcgrad used in amber has the format

```
996
0.002110738107 -0.000907351801 0.001687018358
-0.001089983662 0.000596535645 -0.000837938612
-0.001269634418 0.000426933814 -0.000967084019
-0.000165255832 0.000111137830 0.001623392385
```

Whereas xtb gives the format

```
0.00219465 -0.00095322 0.00174035
-0.00114767 0.00062388 -0.00085454
-0.00131498 0.00044651 -0.00096461
-0.00021916 0.00009229 0.00159509
```

**Describe alternatives you've considered**
I wonder if it is possible to alter the output format or at least have such an option during the compilation.

The way amber read the pc gradient is followed.

```
      ! MM gradients (none if do_grad)
      if ( nclatoms > 0 ) then
        open(iunit, file=basename//'.pcgrad', iostat=ios)
        if ( ios /= 0 ) then
          call sander_bomb('read_results (qm2_extern_orc_module)', &
            'Error opening Orca output file '//basename//'.pcgrad for reading', &
            'Will quit now')
        end if
        do
          read (iunit, '(a)', iostat = ios) read_buffer !First line is nclatoms
          ! End of file; data not found
          if (ios < 0) then
            call sander_bomb('read_results (qm2_extern_orc_module)', &
              'Error reading Orca output from file '//basename//'.pcgrad', &
              '(Empty file)')
          end if
          ! Read MM gradient data
          do i = 1, nclatoms
            read (iunit, '(3(f17.12))') dxyzcl(:,i)
          end do
          exit
        end do
        close(iunit)
      end if
```

**If possible state how you can assist in providing data or code to to implement the feature**
I can test if the xtb is working with amber as desired.

awvwgk:
Writing of the point charge gradient is done here:

https://github.com/grimme-lab/xtb/blob/2609186b7e10f10aa7e7fcd6c1601f207efe553c/xtb/single.f90#L142-L149

Should be fixed by adding a line like
```fortran
write(ich, '(i0)') pcem%n
```

xiki-tempula:
[...]

PR #42, opened 2019-11-06
Change the pcgrad out format to match the amber/orca interface

xiki-tempula:
The output of the gradient of the point charge is not consistent with the orca default output. Thus, when one uses the amber/ORCA interface with xtb. Error occurs. I have changed the pcgrad to the format which is compatible with the amber/ORCA interface and is tested against the ambertools18.

Linked to issue #41 

It is fine that this PR doesn't get merged. I just think it would be nice if people want to do this.

### DIFF
--- a/xtb/single.f90
+++ b/xtb/single.f90
@@ -142,8 +142,9 @@ subroutine singlepoint &
    ! point charge embedding gradient file
    if (pcem%n > 0) then
       call open_file(ich,pcem_grad,'w')
+      write(ich, '(i0)') pcem%n
       do i=1,pcem%n
-         write(ich,'(3f12.8)')pcem%grd(1:3,i)
+         write(ich,'(3f17.12)')pcem%grd(1:3,i)
       enddo
       call close_file(ich)
    endif


awvwgk:
I should check the Orca interface to `xtb` first. If they read the `pcgrad` file from `xtb`, this PR will break compatibility with Orca 4.2.0.

xiki-tempula:
I have checked the ORCA/xtb interface and it seems that ORCA just uses the pcgrad as it is without any read or write, so the compatibility with ORCA4.2.* is fine.

awvwgk:
This PR breaks the xTB/MM interface in Orca 4.2.0, apparently there is a special workaround for the `pcgrad` file produced by `xtb`.

I hate to admit it, but the easiest way is to call it a feature, document it properly and fix the reader on the `amber` side.

xiki-tempula:
@awvwgk Thanks for the checking. I have gone back to check my installation, which appears to be orca 4.2.1. So I guess the xTB/MM interface is probably fixed in Orca 4.2.1, which allows this PR to work. I guess I'm wrong in saying that it works for 4.2.*.

xiki-tempula:
Thank you for the clarification. I guess The orca/xtb interface might crash in other functions not related to the amber/orca/xtb QM/MM functionality. I do agree that this PR will break other programs reading the pcgrad output.

awvwgk:
@xiki-tempula any update here? Do you plan to continue working on this PR?

xiki-tempula:
@awvwgk I have contacted ORCA people and it seems that they plan to fix this question on their end for the next release. I guess this PR can be closed.

awvwgk:
@xiki-tempula thanks for the update.

PR #44, opened 2019-11-07
PDB residue number is not correctly translated to fragments

### LINKED ISSUE #43: PDB residue number is not correctly translated to fragments
### Describe the bug
The new PDB reader expects the residues to be continuous, which is not necessarily the case.

### To Reproduce
Tested with two differently sorted versions of 4qxx, first one has continuous residues, while the second has all hydrogen atoms appended. `xtb` version was 2609186.

```
xtb 4qxx_1.pdb --cycles 1 --opt --gfn 0 --namespace 1
xtb 4qxx_2.pdb --cycles 1 --opt --gfn 0 --namespace 2
```

Input files and output files provided [here](https://github.com/grimme-lab/xtb/files/3818902/4qxx.zip). While the first output file is fine, in the second PDB output the residue numbers do not match the input.

### Expected behaviour
Correctly fragment the structure into the residues. With lesser priority: retain the numbering of the residues.

### Additional context
The residue is saved here while reading:
https://github.com/grimme-lab/xtb/blob/2609186b7e10f10aa7e7fcd6c1601f207efe553c/xtb/molecule_reader.f90#L725-L728

The writer is using this fragment information to determine the residue ID:
https://github.com/grimme-lab/xtb/blob/2609186b7e10f10aa7e7fcd6c1601f207efe553c/xtb/molecule_writer.f90#L295

Maybe one should additionally expand the `pdb_data` label by the residue ID:
https://github.com/grimme-lab/xtb/blob/2609186b7e10f10aa7e7fcd6c1601f207efe553c/xtb/tbdef_molecule.f90#L53-L66

Thanks to @saschmi for reporting, thanks to @liljay42 for looking into this.

liljay42:
Fixes #43


xtb/molecule_reader.f90
```diff
@@ -681,7 +681,7 @@ subroutine read_molecule_pdb(mol, unit, status, iomsg)
    character(len=:), allocatable :: line
    character(len=2) :: a_charge
    integer :: iatom, jatom, iresidue, try, error, atom_type
-   integer :: this_residue, last_residue
+   integer :: this_residue, last_residue, i, j
```
awvwgk: remove `this_residue` and `last_residue` if they are not needed anymore

xtb/molecule_reader.f90
```diff
@@ -735,7 +730,13 @@ subroutine read_molecule_pdb(mol, unit, status, iomsg)
       iomsg = "could not read in coordinates, last line was: '"//line//"'"
       return
    endif
-
+   j=0
+   do i = minval(pdb(:iatom)%residue_number),maxval(pdb(:iatom)%residue_number)
+      if (any(i .eq. pdb(:iatom)%residue_number)) then
+         j=j+1
+         where (i .eq. pdb(:iatom)%residue_number) list = j
```
awvwgk: The logic here would be clearer if we use counter names like `iresidue`/`ifragment`.
Also an `associate` statement could make this more readable:
```fortran
associate(residue_number => pdb(:iatom)%residue_number)
...
end associate
```

liljay42: I tried the associate function but it did not work.

xtb/molecule_reader.f90
```diff
@@ -91,7 +91,7 @@ subroutine read_molecule_xyz(mol, unit, status, iomsg)
 
    read(unit,*,iostat=err) n
    if (err.ne.0) then
-      iomsg = "Could not read number of atoms, check format!"
+      iomsg = "Could not read number of atoms, check format!"100000000000000
```
awvwgk: unrelated change

PR #53, opened 2019-11-22
Do we need a logo?

awvwgk:
This started as a fun project to create a “logo” for `xtb` and was subsequently used on posters which made it kind of official. This PR adds the two `svg` files for the “logo” to the official repository making it really official.

Have a look [here](https://github.com/awvwgk/xtb/tree/xtb-logo/assets/logo) for a depiction of the “logo” and [a way to use them](https://raw.githubusercontent.com/awvwgk/xtb/xtb-logo/assets/logo/README.md) in GitHub flavoured Markdown.

xTB logo alt. xTB logo

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 This is the offical repository of the `xtb` program package developed by the Grimme group in Bonn.
 
+<img src="./assets/logo/xtb.svg" alt="Extended Tight Binding" width="120">
+
 ## Installation
 
 Statically linked binaries (Intel Compiler 17.0.7) can be found at the [latest release page](https://github.com/grimme-lab/xtb/releases/latest).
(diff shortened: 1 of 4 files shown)


lizhiq16:
Since we all use the command line to call xTB, the logo doesn't seem to be necessary... Unless you have plans to compile the windows version, which will bring more users to xtb.
I looked at the xtb logo and thought it was very good. This logo vividly reflects the characteristics of xtb, but it seems that most of the logos are square, and you may need to make some adjustments....

awvwgk:
Thanks for the feedback. I guess the main purpose would be to have something recognizable for posters, talks or web pages.

Making a icon out of it isn't that hard, since it is almost square :).
xTB icon

ghutchis:
IMHO, I'd go with white background and grey edge to the circles, but it's a great idea. (Logos are good for stickers too. :-)

dgasmith:
Recognizable logos are very nice! We would put it on our website and several presentations once we finish integration.

PR #58, opened 2019-11-25
Enable Build with GCC

### LINKED ISSUE #11: Compiling xtb with GCC?
Hi, I'm wondering why you say you cannot build xtb with gcc and there's no plans to support it in the future. Standards compliant code should compile straight away with any compiler. Any issues in the code should be pretty easy to solve...

### LINKED ISSUE #12: Compiling xtb on OSX
*updated and edited*
Compiling on OSX might interest a few people. 

A few hurdles I ran into:
* needs a fully 'shared' build option as static linking is not really a thing on OSX
* requires at least: `meson configure -Db_lundef=false` (possibly also for gcc)

---

current working version on Catalina and Intel19
grab: https://github.com/hokru/xtb/tree/osx3 (shortened meson config plus PR #40)

```
export FC=ifort CC=icc CXX=icpc
meson setup build_intel -Dstatic=false -Db_lundef=false -Db_asneeded=false
```
There will be an error because of wrong linking flags! Ignore and link yourself:
```
ifort  -o xtb 'xtb@exe/xtb_program_main.f90.o' -qopenmp -mkl libxtb.a  -Wl,-rpath,/INSERT_YOUR_PATH/xtb/build_intel/ libxtb.dylib
```

awvwgk:
hopefully fixes #11
also fixes #12 when we are running the CI on OSX!
[Build Status](https://github.com/awvwgk/xtb/actions)
- [x] remove all offending statements such that it compiles with GCC
- [x] check dependencies to allow linking a binary with GCC
- [x] make sure no Intel extensions are used in the code... this gonna be hard :)
  - [x] mostly effects by `-r8` flag (GCC equivalent?)
- [x] make openmp usage compatible with GCC restrictions
  - [x] GCC version >= 8 works
- [x] figure out what is wrong with deferred-length characters and GCC
- [x] get the testsuite running with GCC (for Travis-CI and GH actions)
  - [x] basic GH actions CI :tada:
  - [x] WSC generation is failing (0D always, 3D erratically)
  - [x] xTB Hamiltonians fail erratically
  - [x] atomlist class is somewhat problematic (derived type IO to internal unit)
- [x] have a working binary

This is work in progress and not my highest priority. If you think you can help out here, feel free to open a PR on this PR :)

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -1,0 +1,16 @@
+name: CI
+on: [push, pull_request]
+
+jobs:
+  linux-build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.x'
+    - run: sudo apt install -yq --no-install-recommends gfortran ninja-build liblapack-dev libblas-dev
+    - run: pip3 install meson
+    - run: meson setup build_gcc --buildtype release -Dla_backend=netlib --warnlevel 0
+    - run: ninja -C build_gcc
+    - run: cd build_gcc && meson test --suite basic
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -9,8 +9,18 @@ jobs:
     - uses: actions/setup-python@v1
       with:
         python-version: '3.x'
-    - run: sudo apt install -yq --no-install-recommends gfortran ninja-build liblapack-dev libblas-dev
+    - run: sudo apt install -yq --no-install-recommends gfortran-8 ninja-build liblapack-dev libblas-dev
     - run: pip3 install meson
-    - run: meson setup build_gcc --buildtype release -Dla_backend=netlib --warnlevel 0
-    - run: ninja -C build_gcc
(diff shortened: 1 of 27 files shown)


aslozada:
**atomlist class / gfortran-9**

Hi
An error associated with intent(out) in pure subroutine occurs : elemental pure subroutine  atomlist_defaults.

According to J3/15-007, Fortran  2003  permitted  an  INTENT  (OUT)  argument  of  a  pure  subroutine  to  be  polymorphic;  that  is  not permitted by this part of ISO/IEC 1539. <https://gcc.gnu.org/onlinedocs/gcc-9.1.0/gfortran.pdf>

Change to INOUT solve. Is INOUT condition correct in this case?  

../xtb/tbdef_atomlist.f90:117:48:
  117 | pure elemental subroutine atomlist_defaults(self)
      |                                                1
Error: INTENT(OUT) argument ‘self’ of pure procedure ‘atomlist_defaults’ at (1) may not be polymorphic

awvwgk:
Removing the `pure` attribute is maybe a better idea here. I usually add `pure`/`elemental` when possible, maybe a bit too often.

awvwgk:
Also, setting `intent(inout)` would not to correspond to the wanted behaviour of `atomlist_defaults`, to return a `tb_atomlist` object with default settings.

PR #76, opened 2019-12-09
Preparation for next release (v6.3.0).

awvwgk:
Preparation for a feature preview...

- periodic boundary conditions for GFN1-xTB  and GFN2-xTB Hamiltonians
- API support for periodic GFN1-xTB and GFN2-xTB Hamiltonians (C, Python, ASE)
- general GFN Force Field (GFN-FF) for all elements up to Z=86
- 2D to 3D conversion of SDF input using the GFN-FF
- `crest` support for GFN-FF

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -18,7 +18,7 @@
 # extended tight binding program package
 project('xtb',
         'fortran', 'c', 'cpp',
-        version: '6.2.2',
+        version: '6.3.0',
         license: 'LGPL3',
         meson_version: '>=0.49')
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -21,7 +21,7 @@ with open("README.md", "r") as fh:
     long_description = fh.read()
 
 setup(name="xtb",
-      version="6.2.2",
+      version="6.3.0",
       author="Sebastian Ehlert",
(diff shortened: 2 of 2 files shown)


lizhiq16:
Is there any detailed information (like research paper etc.) about GFN Force Field (GFN-FF)? 
PS.   I think preview 6.3.0 should be 4th releases, but since the  release of 6.2.2, the release number has gone wrong...

sespic:
We are currently working on the GFN-FF paper. We try to publish it as soon as possible, but it will take until next year.

ghutchis:
Is the GFN-FF different from the non-self-consistent "GFN0" method or something new?

awvwgk:
It's indeed a force field, in short it is somewhat similar to the non-selfconsistent GFN0-xTB with the xTB part replaced by force field-like terms. Of course it's something new, at least in the framework of `xtb`.

ghutchis:
Great - I didn't know if GFN0 had been re-named. Thanks and Happy 2020!

ghutchis:
Looks like the GFN-FF paper is now accepted, congrats - the results look very strong: https://onlinelibrary.wiley.com/doi/pdf/10.1002/anie.202004239

Issue #98, opened 2020-01-14
C/C++ Interface Oddities


nabbelbabbel:
Using the C/C++ interface a few odd things have shown, I'd say some of them are bugs.
(If they should be split into separate Issues I don't mind doing that but for now I will gather them here.)

First of all, the interface seem to be working fine overall.  
As a C++ user I have to say: thanks for them!  

The issues encountered are:

### 1. GFN2_calculation Bond Orders
Using 
```c++
extern int
GFN2_calculation(const int* natoms, const int* attyp, const double* charge, 
const double* coord, const SCC_options* opt, const char* output, 
double* energy, double* grad, double* dipole,  double* q, double* dipm, 
double* qp, double* wbo);
```
The `wbo` array is only filled if the `SCC_options` `opt.prlevel` is `>=2`.
The expected behavior would be that the field is always populated, I think.

### 2. GFN1_calculation Bond Orders
Using 
```c++
extern int
GFN1_calculation(const int* natoms, const int* attyp, const double* charge, 
const int* uhf, const double* coord, const SCC_options* opt, const char* output, 
double* energy, double* grad, double* dipole, double* q, double* wbo);
```
The `wbo` array seems to be empty (zeroes) at any given time. 
However, using the executable the `wbo` file is created, hence I assume it should also work using the interface.
Again the expected behavior would be that the field is always populated, I think.

### 3. Parallel Execution of Calculations
3. Working in parallel it is not possible to have a single program run multiple calls to 
`GFN1_calculation` etc. at the same time. The calls do not seem to be threadsafe.
The errors range from thrown exceptions (the timings seem to have some static type members that throw reasonable exceptions) to wild segfaults.
Here, it would be good to note that these calls are not threadsafe and/or actually stop multiple calls vial locks before they go haywire.

### For Reproducibility
Commit: bbcf06e6e92853fe468241d6e4e7bc6cc2642fdf
Compiler: gcc-8.3
[Disclaimer: The code was build using CMake instead of the given toolchain]

awvwgk:
Thanks for the report.

I guess the bond order problem is kind of known (at least I knew that I have to fix it at some point).

A lock would certainly help to prevent several `xtb` calls to run and use the same global variables, but I would rather prefer to purge all global variables from the code base than managing a mutex. So this is a long term goal.

Regarding the CMake build, would you consider sharing the build files? I've never used CMake myself for a project, but there is some interest in building `xtb` with something else than meson.

[...]

PR #99, opened 2020-01-15
CMake build files

nabbelbabbel:
It should now be possible to build XTB using:
```
<cd into repo>
mkdir build
cd build
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/some/place/nice ..
make
make install
```

There are a few things that are unclear to me so the following things are to be done still:
1. <del>The automated generation of `xtb_version.fh`. (Where is the used information read from?)</del> (done)
2. <del>The parameter files need to be added to the install procedure. (Where do they go on install?)</del> (done)
3. <del>The tests need to be added to the build procedure.</del>  (done)
4. <del>Is OpenMP supposed to be optional?</del> (no, but in gcc<8 it is removed)
5. <del>How is the Python part generated?</del> (outside the scope of this PR)
6. <del>This build uses `cmake/Files.cmake` to track existing source files. This list should probably be merged with the one in the meson files, I do not know what the best version of doing that is though.</del> (unwanted)
7. <del>Add installed target files in order to allow for easier linking.</del> (done)


nabbelbabbel:
Here is the first part of the CMake files, I have stated the open questions above.

@awvwgk If you could point me in the right direction with these I think this should be done in no time.

Also It would be good to actually then add the CMake version to the CI, I guess.

awvwgk:
The workflow file for the CI is [here](https://github.com/grimme-lab/xtb/blob/master/.github/workflows/fortran-build.yml). I would prefer to add the CMake build right after the meson build with something like

```yaml
runs: FC=gfortran-8 CC=gcc-8 cmake -B_build -DCMAKE_BUILD_TYPE=Release
runs: make -C _build
runs: ctest _build
```

We could also expand the build matrix for a CMake job, but I would rather not do this, since the GH actions CI really clutters PRs.

awvwgk:
Thanks for your PR. Please keep in mind that the meson build is not the ultimate solution, but the best we currently have. In case we have to modify the meson build as well, we will just do it.

1. For `xtb_version.fh` I am using meson's configure file option, which was “invented” in CMake: https://cmake.org/cmake/help/latest/command/configure_file.html
The syntax `@variable@` seems to be supported by CMake as well.
2. The parameterfiles are installed in `$PREFIX/share/xtb`, see https://mesonbuild.com/Installing.html
3. For a passing testsuite you have to reproduce the `-r8` behaviour of Intel with `-fdefault-real-8` and `-fdefault-double-8` for GCC.
4. OpenMP is the default for `xtb`, the option is supposed to shut it off for CI runs with GCC versions 7 and older, which fail to compile perfectly correct OMP statements (namely `default(none)`).
5. Don't bother about the Python part for now.
6. This might not even be possible due to syntax limitations of either meson or CMake. If we are hosting two build systems here, than we will pay the price to keep both up-to-date.

nabbelbabbel:
> For a passing testsuite you have to reproduce the -r8 behaviour of Intel with -fdefault-real-8 and -fdefault-double-8 for GCC.

I see, but this means I have to ignore like >90 of these warnings?
```
Warning: Use of the NUMERIC_STORAGE_SIZE named constant from intrinsic module ISO_FORTRAN_ENV at (1) is incompatible with option -fdefault-real-8
/fast_scratch/janu/workspace/xtb/xtb/eeq_model.f90:2445:7:

    use iso_fortran_env, wp => real64
       1
```

> The syntax @variable@ seems to be supported by CMake as well.
Yes it does, my question was more about where the actual data is taken from, is this all Git data?

awvwgk:
The warning about incompatible numeric storage size is somewhat annoying, but doesn't break anything. At some point I should switch to `selected_real_kind(1.0d0)` instead.

Currently the used data is partly from `git` and mostly from coreutils programs. So any way to generate this metadata is appropriate here.

.github/workflows/fortran-build.yml
```diff
@@ -14,6 +14,16 @@ jobs:
     - run: FC=gfortran-8 CC=gcc-8 meson setup build_gcc --buildtype release -Dla_backend=netlib --warnlevel 0
     - run: ninja -C build_gcc test
 
+  linux-build-cmake:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: sudo apt install -yq --no-install-recommends gfortran-8 liblapack-dev libblas-dev cmake make
+    - run: mkdir build install
+    - run: cd build && cmake -DCMAKE_INSTALL_PREFIX=../install ..
```
awvwgk:
```yaml
run: cd build && FC=gfortran-8 CC=gcc-8 cmake -DCMAKE_INSTALL_PREFIX=../install ..
```

nabbelbabbel:
So I think this is done from the general CMake side.
Why those tests fail, I have no idea.
Any comments you have are welcome.

FYI: I disabled those tests that expected failure in the CMake tests because I did not see how to convey that expectation to `ctest`.

awvwgk:
First meson supports expected failures and skips (see https://mesonbuild.com/Unit-tests.html), skips return 77 as exit code and are by GNU Standard not failed, but skipped. In case we can't have those, leave them out.

For GFN0-xTB we need the parameter file available, this means the build system has to provide an environment with variables like XTBPATH set correctly to perform the unit tests (reading the parameter file is part of the test suite, of course).

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,178 @@
...
+
+# Populate xtb_version.fh
+set(version ${PROJECT_VERSION})
+execute_process(COMMAND git rev-parse HEAD
+ RESULT_VARIABLE git_return
+ OUTPUT_VARIABLE commit)
+string(REGEX REPLACE "\n$" "" commit ${commit})
+string(TIMESTAMP date "%m/%d/%Y")
+set(author $ENV{USERNAME})
+set(origin $ENV{HOSTNAME})
```
awvwgk: `HOSTNAME` is not always set, for example on my machine and the cluster in Bonn it is always empty.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,178 @@
...
+configure_file("${XTB_ROOT}/xtb/version.f90" "${XTB_ROOT}/xtb/xtb_version.fh" @ONLY)
```
awvwgk: I'm not to keen writing a configured file back to the source directory. I normally have to manage a dozen builds in parallel for the same repository, which becomes fragile if the actual source directories are changing with the build.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,178 @@
...
+)
+set_target_properties(xtb PROPERTIES
+  Fortran_MODULE_DIRECTORY xtb-mod
+  RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}
+)
+target_include_directories(xtb PRIVATE "${XTB_ROOT}/include")
+
+# Install
+install(FILES "${XTB_ROOT}/include/xtb.h"
+        DESTINATION include/xtb/
```
awvwgk: Sure it should be `include/xtb`?

nabbelbabbel: Well I am used to more than just one header file for a program, so that is the default.

.github/workflows/fortran-build.yml
```diff
@@ -14,6 +14,17 @@ jobs:
...
+  linux-build-cmake:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - run: sudo apt install -yq --no-install-recommends gfortran-8 liblapack-dev libblas-dev cmake make
+    - run: mkdir build install
+    - run: cd build && FC=gfortran-8 CC=gcc-8 CXX=g++ cmake -DCMAKE_INSTALL_PREFIX=../install ..
+    - run: cd build && make
+    - run: cd build && make install
+    - run: export XTBHOME=$(pwd) && cd build && make test
```
awvwgk: `XTBPATH=$PWD`

nabbelbabbel:
So there is one failing test left.
```
SHARED_GFN0-xTB_API_PBC
```
Interestingly it only fails in the test executable that is dynamically linked.

.github/workflows/fortran-build.yml
nabbelbabbel: Oh, yes sure

cmake/CMakeLists.txt
awvwgk: meson puts it by default to `include`, but I want consistency, so we should change the meson build instructions. Could you apply this patch?
```patch
diff --git a/meson.build b/meson.build
index 96c1d39..eb185c1 100644
--- a/meson.build
+++ b/meson.build
@@ -20,7 +20,8 @@ project('xtb',
         'fortran', 'c',
         version: '6.2.2',
         license: 'LGPL3',
-        meson_version: '>=0.49')
+        meson_version: '>=0.49',
+        default_options: ['includedir=include/xtb'])
 
 fc = meson.get_compiler('fortran')
 cc = meson.get_compiler('c')
```

cmake/CMakeLists.txt
nabbelbabbel: I will switch it to a CMake variable.

cmake/CMakeLists.txt
nabbelbabbel: Yes, that was a "hack" that was still left.

cmake/CMakeLists.txt
nabbelbabbel: Done

README.md
```diff
@@ -49,6 +52,31 @@ be sourced in your `.bashrc` or `.cshrc` which will make `xtb` and the
 parameter files available. Also a pkg-config file is generated to make
 `xtb` available in other projects as dependency.
 
+### CMake
+
+The CMake build system requires both `make` and `cmake` to be installed, the
+latter has to be version 3.9 or newer.
+
+Building XTB with CMake works with the following chain of commands:
```
awvwgk: XTB -> `xtb`

README.md
```diff
@@ -49,6 +52,31 @@ be sourced in your `.bashrc` or `.cshrc` which will make `xtb` and the
...
+
+```bash
+mkdir build
+cd build
+cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/some/place/nice ..
```
awvwgk: This command *will* be copied and pasted straight to the command line. I would prefer
```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PWD/install
```
or
```
cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$HOME/.local
```

README.md
```diff
@@ -49,6 +52,31 @@ be sourced in your `.bashrc` or `.cshrc` which will make `xtb` and the
...
+make
+make install
+```
+
+The compiled tests can be executed using the following commands:
+
+```bash
+export XTBHOME=$PWD
```
awvwgk: XTBHOME -> XTBPATH
also I don't want to encourage wildly exporting variables. Better set it locally by
```
XTBPATH=$PWD make test
```

.github/workflows/fortran-build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - run: cd build && FC=gfortran-8 CC=gcc-8 CXX=g++ cmake -DCMAKE_INSTALL_PREFIX=../install ..
     - run: cd build && make
     - run: cd build && make install
-    - run: export XTBHOME=$(pwd) && cd build && make test
+    - run: export XTBHOME=$PWD && cd build && make test
```
awvwgk: XTBHOME is deprecated, don't use it.
```
cd build && XTBPATH=$PWD/.. make test
```

nabbelbabbel:
So it seems that the API tests (with gcc) are not stable in parallel.

Setting `OMP_NUM_THREADS=1` in the CI has resolved the problems.

I don't think that this is a particularly nice thing to do but maybe this is related to #98

nabbelbabbel:
I will adapt the README to this `OMP_NUM_THREADS=1` for the tests and then I think this PR is done from my side.

Issue #95, opened 2020-01-10
Importing `interface.py` requires ASE anyway


pierre-24:
**Describe the bug**
I'm trying to achieve [an interface between `gaussian` and `xtb`](https://github.com/pierre-24/g2xtb_interface) in order to perform `xtb:QM` onium calculations. Until recently (since it wasn't possible for me to compile `xtb` in the past), I was using [a custom interface](https://github.com/pierre-24/g2xtb_interface/blob/master/g2xtb_interface/XTB.py) inspired by the previous python file (which does not work with the current version, for this reason), but I would prefer to rely on your implementation of the [`interface.py` file](https://github.com/grimme-lab/xtb/blob/master/python/xtb/interface.py).

I don't really want to  have `ASE` as a dependence for that, since I only need to perform a bridge between the two codes. Not that ASE is a heavy dependence, but it would be weird to require it.

**To Reproduce**
Steps to reproduce the behaviour:

1. I add the `python` directory to `PYTHONPATH`
2. In the python code, I use `from xtb import interface`

The result is, obviously,

```
Traceback (most recent call last):
  File "/home/pbeaujea/g2xtb_interface/interface.py", line 11, in <module>
    from xtb import interface as xtb_interface
  File "/home/pbeaujea/.install_stda/xtb/python/xtb/__init__.py", line 20, in <module>
    from xtb.calculators import GFN1, GFN2, GFN0
  File "/home/pbeaujea/.install_stda/xtb/python/xtb/calculators.py", line 24, in <module>
    from ase.calculators.calculator import Calculator, all_changes
ImportError: No module named 'ase'
```

... Since `__init__.py` is always imported.

**Expected behaviour**

I get why you included `calculators` in  `__init__.py` , but in my case, it is problematic: I could always copy/paste `interface.py` from your repo to mine, but that would not be very elegant (and its probably not the best way due to licencing stuffs).

awvwgk:
Probably the best way would be to separate the Python API from the ASE calculator for this, maybe even in separate repositories. We should discuss this.

pierre-24:
It depends on what you are trying to achieve with it. For me (and in my case), having a `interface.py` file in a `xtb` package is ok, since it is easy to import it this way. The ASE calculator would be an example to show that you can then interface `xtb` with anything, and would be a little bit "higher level abstraction".

But that's my point of view :)

awvwgk:
I thought including dependencies would ensure that they are present downstream:

[...]

PR #106, opened 2020-01-30
Support more file formats for input/output

awvwgk:
Allows `xtb` to read new file formats. Solves #95 on the `xtb` level.

- [x] read and write Gaussian external input files as documented [here](https://gaussian.com/external/)
  - [x] find out default file ending to activate automatic detection (@pierre-24 can I get some input?)
- [x] write Turbomole `lattgrad` file
- [x] write energy/gradient to Orca `engrad` format
- [x] write energy/gradient to Gaussian external format

### DIFF
--- a/scripts/xtb-gaussian
+++ b/scripts/xtb-gaussian
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+2&>1 exec xtb $2 --grad > $4
--- a/scripts/xtb-gaussian
+++ b/scripts/xtb-gaussian
@@ -1,2 +1,2 @@
 #!/usr/bin/env bash
-2&>1 exec xtb $2 --grad > $4
+2>&1 exec xtb $2 --grad > $4
--- a/cmake/Files.cmake
+++ b/cmake/Files.cmake
@@ -55,6 +55,7 @@ set(XTB_SOURCES
   "${XTB_ROOT}/xtb/setparam.f90"
 
   # Header and I/O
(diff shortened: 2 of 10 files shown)


pierre-24:
Here is an example of `*.Ein` and `*.Eou` files for the NH3 molecule: https://gist.github.com/pierre-24/21ebf089b3f2ffa77dd6789175a8a542

pierre-24:
Note that you will still requires a (lighter, tough) interface, since [as you can see in the "script invocation tab"](http://gaussian.com/external/), Gaussian just pass a bunch of different files as input to the program, but thanks anyway ;)

awvwgk:
A small wrapper script like below should suffice:

```bash
#!/usr/bin/env sh
exec xtb $2 --gaussian $3 &> $4
```

pierre-24:
That is also more or less what I had in mind :)

awvwgk:
Huh? A GH error, this is something unusual.

awvwgk:
@pierre-24 Would you mind testing the Gaussian interface? I don't have access to this program.

scripts/xtb-gaussian
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+2&>1 exec xtb $2 --grad > $4
```
jmrd98: There is a typo here, it should be 2>&1.

jmrd98:
I am not @pierre-24 , but I have had a chance to test the interface, and it appears to work superficially. I have not had a chance to explore it in any great detail.

PR #109, opened 2020-02-11
CMake Update

nabbelbabbel:
This PR resolves a problem with the CMake install process if either `git` is not installed on the host machine or the sources have been downloaded and do not include the actual repository data.


nabbelbabbel:
Additionally this resolves some problems when OpenMP is missing, I will un-draft this once I am sure that the CMake part of the PR is done.

nabbelbabbel:
@awvwgk I think this is done.

Issue #158, opened 2020-03-23
MDL and MOL export must correctly annotate bond block tables


tobigithub:
**Describe the bug**
The MDL and SDF export for xtb currently only exports the atom types and connections between atoms, the bond tables are all set to "1" which means single bond. That will lead to very wrong results once people import and re-use optimized MOL and SDF files. Basically once the InchiKey changes from input to output file,  the connections and atom types all subsequent calculations are wrong.

All cheminformatics editors obey the bond table rules that designate for example aromatic or kekule type aromatic. Most of the excellent cheminformatics packages to that too. For example here is an explanation contributed by Robert Belford Professor (Chemistry) at University of Arkansas at Little Rock).  [[LINK]](https://chem.libretexts.org/Courses/University_of_Arkansas_Little_Rock/ChemInformatics_(2017)%3A_Chem_4399%2F%2F5399/2.2%3A_Chemical_Representations_on_Computer%3A_Part_II/2.2.2%3A_Anatomy_of_a_MOL_file) and [[MDL MOL Definitions]](http://c4.cabrillo.edu/404/ctfile.pdf)

![image](https://user-images.githubusercontent.com/2569565/77370576-69d16b00-6d1e-11ea-97be-7a6ae56db02e.png)

As of xtb version 6.3.0 I would not recommend to use any MOL or SDF export, but only XYZ. 
The reason is that SDF/MOL is a very well established format, people will simply assume that there are no issues (there are issues with aromaticity and handling of nitro compounds etc among different packages). However simple bond types always should be correct.

![image](https://user-images.githubusercontent.com/2569565/77372304-aacb7e80-6d22-11ea-9538-7481375df375.png)

The results are not good, the compound is now a completely different compound. This does not happen with other packages that use forcefields, here the bond types are always retained.

![image](https://user-images.githubusercontent.com/2569565/77372363-ccc50100-6d22-11ea-974e-1d2df5e6f323.png)

**To Reproduce**
Run example molecule with single, double, triple and aromatic bond types (2,4) and check output.

Please provide all input and output file such that we confirm your report.

**Expected behaviour**
No change in Inchikey and connection tables, except for few issues (elongated bonds), heteroatoms etc.

**Additional context**

1) One way **for the user to avoid that** currently is to use only XYZ files, because the connection table is missing anyway and has to be restored with openbabel or other tools (xyz2mol). For example  obabel -ixyz -omol result.xyz > result.mol   In many cases the mol file is correct, not in all cases.

[...]

PR #183, opened 2020-04-26
Implement comparison between topologies

### LINKED ISSUE #34: Check and report changes in topology after geometry optimizations
#### Is your feature request related to a problem?
In case a *molfile* or *sdfile* is provided we know the topology of the molecular structure, which could change in a geometry optimization, molecular dynamic or even single point calculations. A change in the topology means either the calculation has yield a spurious result or the input topology was wrong, reporting this at least as warning would help users relying on the connection table found in *molfiles* or *sdfiles* to identify errors.

#### Describe the solution you'd like
After the calculation a complete topology information from the tight-binding wavefunction is available in form of Wiberg bond orders, this could be used to check, update or amend the topology provided from the input.

#### Describe alternatives you've considered
The file `wbo` is written and contains the Wiberg bond orders, it can always be used to amend a topology in a *molfile* or *sdfile* from the caller side.

#### Additional context
Suggested in #26 to aid usability of `xtb` for chemoinformatics.
This requires #30 to be merged.

awvwgk:
discussed in #158
closes #34

For this algorithm we

1. generate a bond topology from the Wiberg bond orders
2. translate the WBO topology to a sorted neighbourlist
3. translate the input topology to a sorted neighbourlist as well
4. compare those neighbourlists w.r.t. length, number of neighbours, actual neighbours and bond orders (stored somewhat hacky in the Wigner–Seitz weight)

This allows to work even on differently sorted bond topologies. In case we fail, or we do not have an input topology (xyz or coord input), we write the new topology as connection table to `xtbtopo.mol` or `xtbtopo.sdf`. The input topology will not be amended by this check.

order | WBO | C-N-O
--- | --- | ---
1 | 0.3 to 1.3 | 0.3 to 1.2
2 | 1.3 to 2.3 | 1.4 to 2.3
3 | greater 2.3 | greater 2.3
4 | | 1.2 to 1.4

@tobigithub @ghutchis Any additional suggestions?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,6 +147,7 @@ list(APPEND srcs
   "${dir}/thermo.f90"
   "${dir}/timing.f90"
   "${dir}/tmgrad.f90"
+  "${dir}/topology.f90"
   "${dir}/wrbas.f90"
   "${dir}/wrgbw.f90"
   "${dir}/wrmodef.f90"
--- a/src/main/property.f90
+++ b/src/main/property.f90
@@ -20,6 +20,7 @@ module xtb_propertyoutput
    use xtb_mctc_io, only : stdout
    use xtb_mctc_symbols, only : toSymbol
    use xtb_cube
+   use xtb_topology
 
 contains
 
@@ -158,6 +159,8 @@ subroutine main_property &
       call print_wbofile(ifile,mol%n,wfx%wbo,0.1_wp)
       call close_file(ifile)
       call print_wiberg(iunit,mol%n,mol%at,wfx%wbo,0.1_wp)
+
+      call checkTopology(iunit, mol, wfx%wbo, 1)
    endif
 
    if (pr_wbofrag) &
--- a/src/meson.build
+++ b/src/meson.build
@@ -145,6 +145,7 @@ srcs += files(
   'thermo.f90',
   'timing.f90',
   'tmgrad.f90',
+  'topology.f90',
   'wrbas.f90',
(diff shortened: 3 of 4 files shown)


tobigithub:
@awvwgk great work, my only input is to add some external validations for the final implementation, or test cases from compounds that do not change bond orders or break bonds during optimization, basically a stable repository of working XYZ2MOL cases. Carbon and hydrogen are probably OK, but elements SNOP sometimes give weird solutions, due to different possible valence states (older molecular editors had problems with exceeding valence states from mol files).
Maybe chargemol (https://sourceforge.net/projects/ddec/files/) is useful.

I am not sure how tricky it is to validate the charge states in final MOL files, because there will be many cases for protomers, deprotomers and other cases. But a major difference will be that MOL files have the total charge state and elemental charges directly stored in the MOL file, so here we need validations of several test cases that this is correctly written by xtb and also imported by xtb (if possible). I have to think about that how to properly test that.

awvwgk:
Here we have to rely on the tight binding wavefunction for efficiency reasons, the populations and bond orders should be okay, but are of course not perfect. One could try to calculate the charge or spin density with `xtb` to get a cube file and feed this to a postprocessing program like Multiwfn, which is of course way more expensive than calculating just WBOs. Another alternative is to localize the MOs with `--lmo` which should give a clearer picture than just bond orders.

Regarding the charge states, we could also validate them. Right now this information is cached and mostly ignored except for the total charge. Maybe worth a separate PR at some point.

PR #205, opened 2020-05-14
xtb (v6.3.0): fixes & improvements

sespic:
Small improvements mainly regarding the optimization routines.


.xtbrc
```diff
@@ -14,7 +14,7 @@ $opt
    maxcycle=0
    maxdispl=2.000000000000000
    hlow=.5000000000000000E-02
-   s6=20.00000000000000
+   s6=100.00000000000000
```
awvwgk, line 17: Sure that this is a save change for all methods?

src/detrotra.f90
```diff
@@ -0,0 +1,72 @@
...
+!
+! xtb is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
+
+subroutine detrotra4(linear,mol,h,eig)
```
awvwgk, line 18: A module would be nice to always have an explicit interface.

.xtbrc
sespic: s6=70

src/detrotra.f90
sespic: now using a module

PR #216, opened 2020-05-18
GFN-FF v1.0.1

sespic:
Adding small improvements to the description of amides:
- scaling of amide N-H acidity parameter
- adjusting amide EEQ charges

### DIFF
--- a/src/gfnff/gfnff_setup.f90
+++ b/src/gfnff/gfnff_setup.f90
@@ -26,7 +26,7 @@ module xtb_gfnff_setup
 
 contains
 
-subroutine gfnff_setup(env,verbose,restart,mol,p_ext_gfnff,gen,param,topo,accuracy)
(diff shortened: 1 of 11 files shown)


src/gfnff/gfnff_setup.f90
```diff
@@ -62,14 +62,14 @@ subroutine gfnff_setup(env,verbose,restart,mol,p_ext_gfnff,gen,param,topo,accura
...
        topo%hbrefgeo=mol%xyz
        if (success) then
           write(env%unit,'(10x,"GFN-FF topology read from file successfully!")')
           return
        else
-          call env%warning("Could not read topology file", source)
+          call env%warning("Could not read topology file, generated new one", source)
           call env%check(exitRun)
           if (exitRun) then
              return
```
awvwgk, line 75: Not necessarily, in case of `--strict` this will turn into an error in the backtrace and the added comment will be confusing.

sespic: done

PR #218, opened 2020-05-19
ICE in GFN-FF implementation

sespic:
Running into a GCC7.3.0 ICE in the new GFN-FF code.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,18 +8,24 @@ addons:
       packages:
          - libblas-dev
          - liblapack-dev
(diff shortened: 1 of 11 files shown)


awvwgk:
Adjusted this branch to also build GCC7 so we avoid regressions once this is fixed.

awvwgk:
Azure pipeline is passing with this changes: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=164276&view=results

PR #244, opened 2020-06-04
Reading charge input

sespic:
Fixed problem with reading charges from command line input


src/gfnff/gfnff_setup.f90
```diff
@@ -208,7 +208,8 @@ subroutine gfnff_input(env, mol, topo)
       ichrg=int(sum(topo%qfrag(1:nf)))
       topo%qfrag(nf+1:mol%n)=9999
     else
-      topo%qfrag=0
+      topo%qfrag(1)=ichrg
```
awvwgk, line 211: Should use `mol%chrg` instead.

PR #265, opened 2020-06-18
Increase default test timeout

susilehtola:
I'm packaging xtb in Fedora, and found out that while the tests run on my workstation, they tend to time out in the build system, leading to build failures.

This patch changes the default test timeout so that the tests don't spuriously fail.


awvwgk:
Good idea, but I would rather add it to `meson_options.txt` and leave the default at 30. You can than overwrite the option in the configure step on the build server.

susilehtola:
Good idea, this should do it.

I don't really understand why, but the tests run in a few seconds when I run `rpmbuild` directly, but some of them take over 10 minutes when I run them in the `mock` chroot(!)

TESTSUITE/meson.build
```diff
@@ -60,92 +60,92 @@ test('Info', xtb_exe, env: xtbenv, args: [
   files('../assets/inputs/xyz/taxol.xyz'),
 ])
```
awvwgk, line 62:
```suggestion

default_test_timeout = get_option('test_timeout')

```

meson_options.txt
```diff
@@ -26,3 +26,5 @@ option('install_modules', type: 'boolean', value: false,
        description: 'Install Fortran module files to include directory.')
 option('build_name', type: 'string', value: 'unknown',
        description: 'Name of the build, will be overwritten automatically by git')
+option('default_test_timeout', type: 'integer', min: 1, value: 30,
```
awvwgk, line 29:
```suggestion
option('test_timeout', type: 'integer', min: 1, value: 30,
```

susilehtola:
Ah yes, I noticed, thanks. I'll rename the option per your suggestion.

Issue #267, opened 2020-06-18
xtb crashes on multiple architectures


susilehtola:
There are crashes in the test suite on multiple architectures. The tests run correctly on

- x86_64
- aarch64
- ppc64le

but they fail on
- i686
- armv7hl
- s390x

Here's are the i686 build logs for reference.

[root.log](https://github.com/grimme-lab/xtb/files/4797942/root.log)
[build.log](https://github.com/grimme-lab/xtb/files/4797943/build.log)

awvwgk:
Thanks for testing, it's nice to know we can run on aarch64 and ppc64le, since we usually only test on x86_64 architectures.

Looks like the problem is a the call to an `external` declared subroutine, so this _might_ go away by adding an explicit interface for `generate_wsc` (or packing it in a module). If it is an implicit interface problem, it will take a significant effort to wrap everything in modules, still this might be worth the effort anyway.

Also, I'm missing the resources and time to test for other architectures than x86_64, if we can come up with a CI solution I'm will try to keep them running once they work, but it's rather unlikely that we (which mean me) will invest significant efforts to get 32 bit or arm architectures running with `xtb`.

susilehtola:
It looks as if there is some memory error.

susilehtola:
fml now there's a circular dependency error. Someone who actually understands the code should have a look at how it should be interfaced properly.

awvwgk:
I guess that was the reason to make it `external` in the first place.

susilehtola:
Well, I already tried to do that:
```
diff --git a/src/type/molecule.f90 b/src/type/molecule.f90
index fe0c6b2..02e2724 100644
--- a/src/type/molecule.f90
+++ b/src/type/molecule.f90
@@ -195,6 +195,14 @@ contains
 !> Constructor for the molecular structure type
 subroutine initMolecule &
      & (mol, at, sym, xyz, chrg, uhf, lattice, pbc)
+
+   interface
+      subroutine generate_wsc(mol,wsc)
+        type(TMolecule), intent(inout) :: mol
+        type(tb_wsc),    intent(inout) :: wsc
+      end subroutine generate_wsc
+   end interface
+
    type(TMolecule), intent(out) :: mol
    integer, intent(in) :: at(:)
    character(len=*), intent(in) :: sym(:)
```
and here's what I get:
```
[  0%] Building Fortran object CMakeFiles/xtb-object.dir/src/type/molecule.f90.o
[...]

PR #268, opened 2020-06-18
WIP: Modularize generate_wsc

susilehtola:
Work to solve #267


susilehtola:
This is not going to be a solution due to circular dependencies.

PR #269, opened 2020-06-18
Add interfaces to generate_wsc

susilehtola:
The patch adds explicit interfaces for `generate_wsc`, as well as fixes a bug in `generate_wsc` that causes test failures on a number of architectures.

### DIFF
--- a/src/generate_wsc.f90
+++ b/src/generate_wsc.f90
@@ -90,7 +90,7 @@ subroutine generate_wsc(mol,wsc)
             end do
          end do
          ! sanity check; otherwise code below crashes sometimes
-         if ( c .eq. 0 ) cycle
+         if (c .eq. 0) cycle
          ! get first image with same dist
          ! find minimum in dist-array and assign it to minpos = minimum position
          trans=.true.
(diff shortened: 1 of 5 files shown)


susilehtola:
Clean koji build on all architectures
https://koji.fedoraproject.org/koji/taskinfo?taskID=45854116

src/generate_wsc.f90
```diff
@@ -89,6 +89,8 @@ subroutine generate_wsc(mol,wsc)
                end do
             end do
          end do
+         ! sanity check; otherwise code below crashes sometimes
+         if ( c .eq. 0 ) cycle
```
awvwgk, line 93: Not a fan of too much whitespace
```suggestion
         if (c .eq. 0) cycle
```

src/gfnff/calculator.f90
```diff
@@ -36,6 +37,13 @@ module xtb_gfnff_calculator
    use xtb_gfnff_generator, only : TGFFGenerator
    use xtb_gfnff_eg
    implicit none
+   interface
+      subroutine generate_wsc(mol,wsc)
+        import :: TMolecule, tb_wsc
+        type(TMolecule), intent(inout) :: mol
+        type(tb_wsc),    intent(inout) :: wsc
+      end subroutine generate_wsc
+   end interface
```
awvwgk, line 46: I know we are not consistent with the indentation everywhere, but the preference is 3 spaces.

PR #270, opened 2020-06-18
Flush output at end of write functions

susilehtola:
Output files aren't currently flushed, which makes it painstaking to follow progress of the optimization in xtb.

This PR adds `flush(unit)` at the ends of writer functions, allowing one to visualize e.g. `xtbopt.log` with Jmol or other programs.

### DIFF
--- a/src/io/writer.f90
+++ b/src/io/writer.f90
@@ -88,6 +88,8 @@ subroutine writeMolecule(self, unit, format, energy, gnorm, number)
       endif
    end select
 
+   ! Flush file so that the output file can be visualized during optimization
+   flush(unit)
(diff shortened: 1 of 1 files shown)


awvwgk:
It's might be easier to flush the unit in the wrapper rather than the actual writer implementations, this would ensure the unit is flushed even if new trajectory formats are added. See: https://github.com/grimme-lab/xtb/blob/master/src/io/writer.f90

Not sure about the non-geometry outputs, as they are usually not used in trajectory-type output, but it probably won't hurt.

susilehtola:
you're right :clown_face:

PR #274, opened 2020-06-23
Add COSMO implicit solvation model

awvwgk:
- use ddCOSMO library to solve COSMO equations


ghutchis:
Looks great! Is it possible to get the solvent volume / surface area printed as computed properties (or via the Python API). I'd love to use XTB / CREST to augment solubility predictions.

awvwgk:
See #336 for the extension of the C-API w.r.t. to solvation. I have not yet decided on the necessary API objects and calls to implement this, but it is on the list.

ghutchis:
I also saw in the documentation about:
$write
   gbsa=true

This looks like it would give what we want - grabbing the SASA (even h-bond terms).

awvwgk:
@ghutchis I always advise against parsing standard output from `xtb`, save yourself the pain to decrypt the ascii table and just add the terms you are interested in to the [JSON dump](https://github.com/grimme-lab/xtb/blob/master/src/main/json.f90).
Also, we should probably move this discussion to a new issue than...

PR #288, opened 2020-06-28
Add build instructions with cygwin

njustcodingjs:
Add the way to get Cygwin-build xtb and upload the cygwin release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,18 @@ It is possible to list all of the versions of `xtb` available on your platform w
 ```
 conda search xtb --channel conda-forge
 ```
+### Cygwin-build xtb
+Now Windows users can build 'xtb' use 'Cygwin' and 'cmake'. The process is similar to the Cmake build.
+```bash
+mkdir build
+pushd build
+cmake /STACK:4000000000 -DCMAKE_BUILD_TYPE=Release ..
+make
+ctest
+popd
+```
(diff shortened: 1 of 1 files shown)


njustcodingjs:
I think this is a possible way to build xtb under Unix-like system. The '/STACK' parameter was used because I cannot use the command like 'ulimt -s unlimited' under the Windows system. I have to give enough stack to the program when I build it. In Unix-like system, if you have 'bash', this paramter is not necessary. 'xtb' can be built in the same way by using Cmake.

awvwgk:
Okay, after some reading on cygwin things are getting clearer for me. Since you said all tests are passing, it is also able to handle Unix like file paths, which should remove the concern about Windows vs Unix path separators.

Therefore, I would ask you to move the cygwin section to the CMake guide (https://github.com/grimme-lab/xtb/blob/master/cmake/README.adoc), as the guide there is more suitable than the short overview in the main README.

Since you mentioned distribution, I would handle this similar to OSX, meaning there will be no binary release on GitHub, but we will aim for a conda-forge package.

njustcodingjs:
OK, I will refork the repository. These days I tried to install a macos virtual machine on windows. If there is any progress, I will inform you in time.

awvwgk:
Closed in favour of #296

PR #294, opened 2020-06-29
Added alternative dumpstep flags for $md block

pprcht:
The `mddump` keyword only existed in the legacy `$set` block but is required for the metadynamics reference structre and  `scoord` file dump. It should now work via the `sdump` or `dumpcoord` keyword in the `$md` block. Also, to distinguish it from the `dump` keyword (for writing to the `xtb.trj` file) two alternative keywords `dumpxyz` and `dumptrj` are added.

Furthermore a `--ceasefiles` flag was added to reduce the amount of files written by `xtb`, which might be useful for I/O within `crest`.

### DIFF
--- a/src/set_module.f90
+++ b/src/set_module.f90
@@ -1020,6 +1020,14 @@ subroutine set_enso_mode
    enso_mode = .true.
 end subroutine set_enso_mode
 
+subroutine set_ceasefiles(env)
+   implicit none
+   type(TEnvironment), intent(inout) :: env
+   ceasefiles = .true.
+   call set_write(env,'wiberg','false')
+   call set_write(env,'charges','false')
+end subroutine set_ceasefiles
+
 subroutine set_samerand
    implicit none
    samerand = .true.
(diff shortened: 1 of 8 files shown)


src/set_module.f90
```diff
@@ -2146,13 +2159,18 @@ subroutine set_metadyn(env,key,val)
...
    select case(key)
    case default ! do nothing
       call env%warning("the key '"//key//"' is not recognized by metadyn",source)
    case('save')
       if (getValue(env,val,idum).and.set1) metaset%maxsave = idum
       set1 = .false.
+   case('dump')
+      call set_siman(env,'dump',val)
+      if (getValue(env,val,ddum).and.set5) dump_md = ddum
+      set5 = .false.
```
awvwgk, line 2173: This is probably not a good idea, as it allows to overwrite `dump_md` even if it should be locked already through `set_md`. Let's remove this part and also remove the `set_siman` completely, such that it is only set in the `set_md` block.

src/set_module.f90
```diff
@@ -1020,6 +1020,14 @@ subroutine set_enso_mode
    enso_mode = .true.
 end subroutine set_enso_mode
 
+subroutine set_ceasefiles(env)
+   implicit none
+   type(TEnvironment), intent(inout) :: env
+   ceasefiles = .true.
+   call set_write(env,'wiberg','false')
+   call set_write(env,'charges','false')
```
awvwgk, line 1028: Those should probably be in `main.F90` as well

PR #296, opened 2020-07-03
Update README.adoc

njustcodingjs:
Add the Cygwin-build xtb in README.adoc

### DIFF
--- a/cmake/README.adoc
+++ b/cmake/README.adoc
@@ -126,3 +126,16 @@ ninja -C _build install
 
 If you have chosen make as generator use the same command with make.
 Depending on the installation prefix and your user rights ninja/make might ask for the `root` access to perform the installation.
+
+## Cygwin-build xtb
+Now Windows users can build 'xtb' use 'Cygwin' and 'Cmake'. The process is similar to the Cmake build. This build has been tested under GNU 7.4.0 compiler.
+```bash
+mkdir build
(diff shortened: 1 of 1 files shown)


awvwgk:
Thanks, I currently trying to setup a cygwin build on appveyor (see https://ci.appveyor.com/project/awvwgk/xtb), but so far this is not really successful. Maybe I'm missing some packages, so it might be a good idea to have a list of required cygwin packages in the guide as well.

njustcodingjs:
Yes, it seems to be lack of a blas library.

njustcodingjs:
GNU compiler and math library should be added when installing cygwin. I will check the version this night.

awvwgk:
Looks like I got the build working on the Appveyor server, both for CMake and meson, yet the testsuite is not running correctly and therefore the produced binary invalid. This is the list of cygwin packages used so far:

build | `--packages`
--- | ---
CMake | `python3-devel,gcc-fortran,cmake,make,liblapack-devel,liblapack,libopenblas`
meson | `python3-devel,gcc-fortran,meson,ninja,liblapack-devel,liblapack,libopenblas`

One thing I noticed, the build is just incredibly slow, taking 10 min with CMake/make and 7 min with meson/ninja and this is just the build not the actual testsuite. Seems like I can't really parallelize the build, which is not good.

PR #311, opened 2020-08-05
CMake support for library suffix

susilehtola:
Allow use of library suffix on distributions that rely on it, e.g. `lib64` on Fedora and Red Hat, and `lib32` on Debian and Ubuntu.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,8 +105,8 @@ install(FILES
 )
 install(TARGETS lib-xtb-static lib-xtb-shared xtb-exe
   EXPORT xtbTargets
-  LIBRARY DESTINATION lib
-  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib${LIB_SUFFIX}
+  ARCHIVE DESTINATION lib${LIB_SUFFIX}
   RUNTIME DESTINATION bin
 )
 # CMake package files


susilehtola:
> Means `-DLIB_SUFFIX=64` has to be set to get the correct suffix, right? And if I set `-DLIB_SUFFIX=/pkgconfig` it even ends up in another directory...

Yes, but this done automatically in e.g. Fedora as you can see from ```rpm -E %{cmake}``` which contains
```
  /usr/bin/cmake \
         \
         \
        -DCMAKE_C_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_CXX_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_Fortran_FLAGS_RELEASE:STRING="-DNDEBUG" \
        -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON \
        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
        -DSHARE_INSTALL_PREFIX:PATH=/usr/share \
%if "lib64" == "lib64" 
        -DLIB_SUFFIX=64 \
%endif 
        -DBUILD_SHARED_LIBS:BOOL=ON
```

> Maybe `include(GNUInstallDirs)` is the better option here? See https://cmake.org/cmake/help/v3.18/module/GNUInstallDirs.html

Maybe.

awvwgk:
Currently the CMake build does only offer minimal features compared to meson as I consider meson the preferred way to build `xtb`. I will have a detailed look about the CMake build and see if I can make it more consistent if the meson one.

The CMake build is currently missing a lot:
- man pages are not build nor installed
- pkg-config support
- TCL environment module support

PR #312, opened 2020-08-05
Use soname in CMake builds

susilehtola:
The meson builds set a soname of `.6`; this PR makes CMake set a soname as well.


awvwgk:
Implemented in #318

PR #313, opened 2020-08-05
Same header include directory in CMake as in meson.

susilehtola:
CMake is placing the `xtb.h` header in `include/xtb/` whereas Meson installs them in `include/`; with this PR both install the header in `include/`.


awvwgk:
Implemented with `GNUInstallDir` in #318.

PR #316, opened 2020-08-05
Fix crash in get_environment_variable()

susilehtola:
This PR solves a long-standing issue in xtb, mysterious crashes on the build systems in the check phase. After several hours of debugging, I've finally found the root problem.

The mysterious crashes were always due to `Fortran runtime error: Zero-length string passed as value to get_environment_variable.`

The problem is triggered whenever you read the value for a variable that is not set. In this case the length is 0, which is properly allocated, but then the GNU Fortran library crashes when the variable is read since the result array has zero length.

The problem was triggered in the tests, since neither CMake nor Meson sets `XTBHOME`.

### DIFF
--- a/src/mctc/systools.F90
+++ b/src/mctc/systools.F90
@@ -155,13 +155,17 @@ subroutine rdvar(name,var,iostat)
          call raise('E','could not be allocated',1)
       endif
    endif
-   call get_environment_variable(name,var,status=err)
-   if (err.ne.0) then
-      if (present(iostat)) then
-         iostat = err
-         return
-      else
-         call raise('E','System variable corrupted',1)
+   ! If the environment variable has not been set, l=0, and the
+   ! following get_environment_variable call crashes.
+   if (l.gt.0) then
+      call get_environment_variable(name,var,status=err)
+      if (err.ne.0) then
+         if (present(iostat)) then
+            iostat = err
+            return
+         else
+            call raise('E','System variable corrupted',1)
+         endif
       endif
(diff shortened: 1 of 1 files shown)


awvwgk:
Could I get some details on the platform, operation system and compiler you are running into this issue? Because it seems to be not a problem in the CI.

awvwgk:
Seems like this was source to a bug in Intel Fortran as well some time ago:
https://community.intel.com/t5/Intel-Fortran-Compiler/get-environment-variable-bug/td-p/1129490

@susilehtola I found your report at the GCC bug tracker (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96486), let's wait for them to confirm before we move on with a workaround here.

susilehtola:
> Could I get some details on the platform, operation system and compiler you are running into this issue? Because it seems to be not a problem in the CI testing.

This is on the Fedora buildsystem, which made it a pain in the ass to debug.

I guess at the end the issue was caused by `HOSTNAME` being an empty string in the check phase. Applying this and #317 leads to more informative error messages.

susilehtola:
But this still doesn't resolve the crash you get when e.g. `HOSTNAME` is an empty string.

awvwgk:
> But this still doesn't resolve the crash you get when e.g. `HOSTNAME` is an empty string.

I just checked locally with three versions of GCC (7.5, 8.4 and 10.1) and it works fine for

```
HOSTNAME= xtb --version
```

PR #317, opened 2020-08-05
Add missing sanity checks for environment variables

susilehtola:
This PR addresses weird crashes that occur related to #316, when there are environment  variables that haven't been set.

### DIFF
--- a/src/type/environment.f90
+++ b/src/type/environment.f90
@@ -135,8 +135,17 @@ subroutine initEnvironment(self, strict)
 
    call rdarg(0, self%whoami, err)
    call rdvar('HOSTNAME', self%hostname, err)
+   if (err /= 0 .or. len(self%hostname) <= 0) then
+     call raise('E',"Parsing HOSTNAME failed!",1)
+   end if
    call rdvar('HOME', self%home, err)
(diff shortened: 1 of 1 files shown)


src/type/environment.f90
```diff
@@ -135,8 +135,17 @@ subroutine initEnvironment(self, strict)
 
    call rdarg(0, self%whoami, err)
    call rdvar('HOSTNAME', self%hostname, err)
+   if (err /= 0 .or. len(self%hostname) <= 0) then
+     call raise('E',"Parsing HOSTNAME failed!",1)
+   end if
```
awvwgk, line 140: We can't do this here for multiple reasons:
1. the error is not handled on purpose, because it is only a minor inconvenience to not have the name of the host
2. we rely on the implementation to not allocate the deferred-length character variable and to not kill the application, we will use the allocation state later to check whether is was set or not
3. since the deferred-length character might not be allocated, and we don't have short-circuit evaluation in Fortran the check with `len` intrinsic on an unallocated deferred-length character variable can result in an invalid memory access
4. we can't raise an error in the initialization of the error handler responsible for handling the error

awvwgk:
I think this is resolved now the additional check for the length of the environment variable to retrieve.

PR #323, opened 2020-08-07
Read dipole moments for mopac interface calculations

pprcht:
Signed-off-by: Philipp Pracht <pracht@thch.uni-bonn.de>

### DIFF
--- a/src/extern/mopac.f90
+++ b/src/extern/mopac.f90
@@ -146,7 +146,7 @@ subroutine checkMopac(env)
 end subroutine checkMopac
 
 
-subroutine runMopac(env,nat,at,xyz,energy,gradient)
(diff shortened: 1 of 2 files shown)


src/extern/mopac.f90
```diff
@@ -210,6 +211,11 @@ subroutine runMopac(env,nat,at,xyz,energy,gradient)
             energy = dum(num)*evtoau
             cycle read_mopac_output
          endif
+         if (index(line,'DIP_VEC:DEBYE') > 0)then
+            call readl(line,dum,num)
+            dipole(1:3) = dum(2:4)
```
awvwgk, line 216: Requires conversion from Debye to atomic units.

PR #330, opened 2020-08-20
Reworked RMSD bias in MD simulations (static vs dynamic)

pprcht:
Some minor rework for RMSD bias handling in MD simulations.
If the RMSD bias has been initialized for an existing ensemble it is now possible to use the bias as statical potential without any modification during the MD.
Otherwise the RMSD bias will work dynamically, as before.


awvwgk (approved):
Thanks, ready to be merged.

PR #333, opened 2020-08-25
Error in GFN-FF torsion potential

sespic:
- Implementation error within the force constant initialization routine for the torsion potential in GFN-FF version 1.0.1
- The following are affected: Amide bonds, Carbonyl-Cα bonds


awvwgk (commented):
Bumping the GFN-FF version to 1.0.2 seems prudent.

PR #335, opened 2020-09-01
Implemented ALPB default parameters!

MtoLStoN:
Added the default parameters for the newly implemented ALPB solvation model.


src/solv/model.f90
```diff
@@ -210,48 +234,149 @@ subroutine loadInternalParam(self, env, solvent, level)
...
-      end select
-   end if
-
-   if (.not.allocated(param)) then
-      call env%error("solvent: '"//solvent//"' is not parametrized", source)
-      return
-   end if
+   select case(self%kernel)
 
+  case (gbKernel%still)
```
awvwgk, line 239: Please fix the indentation here, one indentation level is three spaces here, the block inside the `case` should be indented twice (six spaces).

src/solv/model.f90
```diff
@@ -210,48 +234,149 @@ subroutine loadInternalParam(self, env, solvent, level)
...
+    case('octanol');      param = gfn1_alpb_octanol
+    case('woctanol');      param = gfn1_alpb_woctanol
+    case('phenol');      param = gfn1_alpb_phenol 
+    case('ch2cl2','dichlormethane'); param = gfn1_alpb_ch2cl2
+    case('chcl3','chloroform');      param = gfn1_alpb_chcl3
+    case('cs2');          param = gfn1_alpb_cs2
+    case('dmso');         param = gfn1_alpb_dmso
+    case('ether');        param = gfn1_alpb_ether
+    case('h2o','water');  param = gfn1_alpb_water
+    case('methanol');     param = gfn1_methanol !still gbsa parameter
```
awvwgk, line 365: In this case, make it explicit than:
```suggestion
    case('methanol')
       self%paramFile = 'internal GFN1-xTB/GBSA'
       param = gfn1_methanol !still gbsa parameter
```

awvwgk (approved):
Looks good now.

Issue #322, opened 2020-08-07
Error termination for charged molecules using GFN-FF (xTB version 6.3.2)


NicolaiRee:
### xTB error termination when performing calculations on charged molecules using GFN-FF (xTB version 6.3.2).
**This works with GFN-X (X=0,1,2)**

#### xTB version:
xtb version 6.3.2 (954f15c) compiled by 'conda@5a45a0871d67' on 2020-07-02

#### Command-line input:
`xtb --gfnff file.xyz --opt --chrg 1 --uhf 0`
or
`xtb --gfnff file.sdf --opt --chrg 1 --uhf 0`

(file.xyz and file.sdf are given below)

#### xTB terminations:
```
[ERROR] Program stopped due to fatal error
-3- Single point calculation terminated
-2- gfnff_calculator_singlepoint: Force-field method terminated
-1- gfnff_eg: EEQ charge constrain error
```

**However, if I e.g. start by running a GFN-1 calculation using file.sdf as input and then run a GFN-FF calculation also with file.sdf as input, then everything works because the program uses some of the files created by the GFN-1 calculation.**

### Additionally 
**file.xyz**
```
13

O          1.46380       -2.28250       -0.03700
C          0.83720       -1.22990       -0.01210
C          1.48320        0.05370        0.23580
C          0.78320        1.22590        0.25700
C         -0.58340        1.21710        0.04530
C         -1.29780       -0.01700       -0.19940
N         -0.49180       -1.19440       -0.21430
H          2.55640        0.03220        0.40160
H          1.29720        2.17180        0.44200
H         -1.15950        2.13480        0.05850
H         -1.83570        0.07420       -1.15500
H         -2.08310       -0.11920        0.56530
H         -0.96940       -2.06670       -0.38760
```

**file.sdf**
```

  xtb     08072014173D
xtb: 6.3.2 (954f15c)
 13 13  0     0  0            999 V2000
    1.4896   -2.2438   -0.0275 O   0  0  0  0  0  0  0  0  0  0  0  0
    0.8475   -1.2058   -0.0075 C   0  0  0  0  0  0  0  0  0  0  0  0
    1.4981    0.0466    0.2360 C   0  0  0  0  0  0  0  0  0  0  0  0
    0.7744    1.2240    0.2564 C   0  0  0  0  0  0  0  0  0  0  0  0
   -0.5681    1.2354    0.0512 C   0  0  0  0  0  0  0  0  0  0  0  0
[...]

PR #341, opened 2020-09-04
Bug fix: Initialize qfrag for SDF and mol case in gfnff_setup  Resolves: #322

Thomas3R:
Added the initialization of the fragment charge qfrag for the sdf and mol file case as proposed by @awvwgk .
This seems to resolve the problem. I tested it for the given file.sdf, a self made molfile and for some few charges each.

### DIFF
--- a/src/gfnff/gfnff_setup.f90
+++ b/src/gfnff/gfnff_setup.f90
@@ -152,9 +152,6 @@ subroutine gfnff_input(env, mol, topo)
     end do
     ichrg=idint(sum(topo%qfrag(1:topo%nfrag)))
     write(env%unit,'(10x,"charge from pdb residues: ",i0)') ichrg
-    ! initialize qfrag as in the default case
-    topo%qfrag(1)=mol%chrg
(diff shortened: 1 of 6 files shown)


src/gfnff/gfnff_setup.f90
```diff
@@ -152,6 +152,9 @@ subroutine gfnff_input(env, mol, topo)
     end do
     ichrg=idint(sum(topo%qfrag(1:topo%nfrag)))
     write(env%unit,'(10x,"charge from pdb residues: ",i0)') ichrg
+    ! initialize qfrag as in the default case
+    topo%qfrag(1)=mol%chrg
+    topo%qfrag(2:mol%n)=0
```
awvwgk, line 157: This is overwriting the PDB case, please don't do this.

awvwgk (changes requested):
We should add a regression test for this as well. I'll try to come up with a simple one.

awvwgk (approved):
Thanks, looks good now.

PR #342, opened 2020-09-06
Initial GPU port of xtb

patrickatkinson:
This pull request adds initial GPU support to xtb using OpenACC and cuSOLVER. Routines that now have GPU support are:

- `build_SDQH0`
- `d4_gradient`
- eigenvalue solvers

GPU and cuSOLVER support can be enabled/disabled through meson build options (`meson_options.txt`). All tests pass except  those that were previously failing with the NVIDIA/PGI compilers (tested with PGI 19.10 and NVFORTRAN 20.7).


awvwgk:
Thanks I will have a look ASAP

Issue #345, opened 2020-09-12
Allow changing of time constant for MTD bias


awvwgk:
**Is your feature request related to a problem? Please describe.**
The time constant for the RMSD bias in meta dynamics is currently set as magic constant to 0.03, see here:

https://github.com/grimme-lab/xtb/blob/5880421ef2c6a31b67f8177336fcb9ac40f8209a/src/dynamic.f90#L429-L430

**Describe the solution you'd like**
The time constant should be read from the detailed input:

```
$metadyn
   bias_ramp_time=0.03
```

PR #346, opened 2020-09-12
Allow MTD bias time constant adjustment via detailed input

pprcht:
#345 

Signed-off-by: Philipp Pracht <pracht@thch.uni-bonn.de>


awvwgk (approved):
Thanks, looks good.

Note: Intel build with meson is broken since the oneAPI beta08 -> beta09 update. So don't mind the failing Intel build.

PR #353, opened 2020-09-21
GPU parallelization in DFT-D3, GBSA, Born, repulsion and CN

jbstueckrath:
GPU implementation of some pre-SCF routines


awvwgk:
Currently fails with

```
Failing in Thread:1
call to cuStreamSynchronize returned error 700: Illegal address during kernel execution
```

PR #362, opened 2020-10-04
Fixes incorrect basename detection when path does not contain '/'

clavigne:
As of now, if the initial geometry file is given as a relative path without any / in it, the basename is empty. For example, running 

    xtb Gau-1234.EIn

generates an output file at `.EOu` as opposed to the correct `Gau-1234.EOu`, because the basename is not detected and defaults to the empty string. Running

    xtb ./Gau-1234.EIn

works fine however.

This very small PR fixes this by explicitly setting the position of the / (when it's not there) to just before the beginning of the filename (pos 0).

### DIFF
--- a/src/mctc/filetypes.f90
+++ b/src/mctc/filetypes.f90
@@ -156,6 +156,16 @@ subroutine generateFileMetaInfo(Name, path, basename, extension)
    iDot = index(name, '.', back=.true.)
    iSlash = index(name, '/', back=.true.)
 
+   ! When given a name like "Gau-3333.Ein", the basename should be "Gau-3333",
+   ! so if there is no '/', we assume there is an implicit './' at the
+   ! beginning of name.
+   if (iSlash > 0) then
+      path = name(:iSlash)
+   else
+      path = ''
+      iSlash = 0
+   endif
+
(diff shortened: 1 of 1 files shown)


src/mctc/filetypes.f90
```diff
@@ -156,24 +156,29 @@ subroutine generateFileMetaInfo(Name, path, basename, extension)
...
    iSlash = index(name, '/', back=.true.)
 
+   ! When given a name like "Gau-3333.Ein", the basename should be "Gau-3333",
+   ! so if there is no '/', we assume there is an implicit './' at the
+   ! beginning of name.
+   if (iSlash > 0) then
+      path = name(:iSlash)
+   else
+      path = ''
+      iSlash = 0
```
awvwgk, line 166: Already zero at this point, no need to set it again

awvwgk (commented):
Thanks, your right, the check should be independent of the value of `iSlash`, the only necessary change is to remove the check for `iSlash` being greater zero to make this work.

Also, could you sign off your commit.

clavigne:
Thanks, will sign off and remove extraneous assignment.

clavigne:
Made suggested changes, thanks

awvwgk (approved):

Issue #329, opened 2020-08-18
Gaussian / xtb interface that provides Hessians


clavigne:
xtb currently ships with a preliminary interface to Gaussian. However, the basic interface doesn't include the ability to pass Hessian information, which is critical for some optimization problems (such as finding transition states), and can be used to improve the speed of Gaussian's own ab initio optimizations by reading xtb Hessians as initial Hessian guesses. In addition, the current xtb-gaussian interface is a bit fragile with respect to paths for input-output files.

For work we are currently engaged in, I coded a basic single-file Perl script that addresses these problems (based on an earlier Python version by @kjelljorner and @gabegomes ). It's located [here](https://github.com/aspuru-guzik-group/xtb-gaussian), it's pretty robust and quite fast. It doesn't use any external perl packages so it should just work out of the box.

I currently have it licensed under MIT, meaning you can just integrate it directly in xtb if you want. I considered submitting it as a PR but I wasn't sure what your opinions are on introducing Perl code or on interop with Gaussian. If you are interested, I can probably write the necessary fortran code to move the feature into xtb proper, but I'm not sure when I will have time to do so.

Thanks for the great program!

awvwgk:
I guess, I start with my usual Gaussian disclaimer. Since I don't have access to Gaussian, I can't test any compatibility features.™

Leaving that aside, thanks for writing this up. Not sure if I want to include something I can't officially support, but I think a link from the documentation, similar to the Orca interface is possible.

Also, I would prefer to fix issues in the `xtb` codebase first before working around them in a wrapper script, looks like you identified some. Most should be straight-forward to implement in `xtb` leaving the wrapper script with the job of reshuffling the commandline arguments.

The information used to infer whether Gaussian specific output should be written or not is currently stored in the input file type and accessible everywhere the molecular structure is available:

https://github.com/grimme-lab/xtb/blob/0a0ed2ccddcf41fb423b90d6f16439bc68be4580/src/prog/main.F90#L800-L809

clavigne:
Oh that actually looks way easier than I thought, thanks for the pointer. I guess I can just match the turbomole code by adding fres to writeResultsGaussianExternal? 

That's probably what I should have done from the get go :D I'll PR it when it's ready and tested.

awvwgk:
For the hessian it dependents on what Gaussian is expecting, a non-massweighted unprojected hessian is written here:
[...]

PR #363, opened 2020-10-05
Adds output of Hessians to xtb when acting as an external program for Gaussian 16

clavigne:
This is a feature improvement related to #329 .

Basically, I've added Hessian output capability for xtb acting as an external program invoked by Gaussian 16. I've made the changes to `writeResultsGaussianExternal` that are necesserary for the Hessian output.

This would have been otherwise extremely straightforward, except for the fact that the raw Hessian matrix (before mass-weighting, which is what Gaussian expects) is not saved in `numhess` for (I assume) memory usage reasons. I didn't want to just have the code keep the raw Hessian every time for the vast majority of users not running under Gaussian, so I short-circuited `numhess` instead to immediately return when the input file is a Gaussian external file. (This additionally has the advantage that it avoids an extraneous Hessian diagonalization by xtb, which gets expensive for large molecules.) 

This solution is far from elegant as it required adding some more global state. Feel free to suggest an alternative.


clavigne:
In addition, I'd be very happy if there was an "easy" way to control runtype based on the `mode` flag in
https://github.com/grimme-lab/xtb/blob/7e8e21c2768696ed49aa41acf331cfdb22d91bae/src/io/reader/gaussian.f90#L47
which is set by Gaussian to 1 for gradient and 2 for Hessian evaluation. I've considered calling `set_runtyp` in the reading subroutine but that's maybe a bit too much global state manipulation. The alternative I can see is to create an entirely new "under gaussian control" runtype.

Very interested in your opinion on this, thanks.

awvwgk:
The read mode could be cached in the molecular structure type and the main routine can than decide if it is able to invoke `set_runtyp`. There is no guarantee that this will work if the user already provided a flag, which locks this, a safer way would be to allow the xtb-gaussian wrapper to query the first line of the input and set the flags accordingly:
```bash
flag=$(awk 'NR == 1 {if ($2 == 2) {print "--hess"} else {print "--grad"} }' $2)
```

clavigne:
Thanks for this and you are right. awk is the right tool to use here and will allow me to remove any final Perl dependencies, moving the whole thing to a self contained two line Bash script. 

Will update (and sign off commits! Sorry again)

Should I add a unit test too? Without Gaussian of course, just testing file io structure.

awvwgk:
We don't have unit tests for the frequency calculation yet, since it is a bit costly in the CI. I'll have a look myself while checking the PR.

clavigne:
I made all the suggested changes. The script with awk works beautifully, and the performance is much better than by passing through perl or python to do the required Hessian reformatting.

Thanks for your help!

awvwgk:
Overall it still looks a bit hacky, but as long as it only triggers with Gaussian input we can be sure it won't mess up regular `xtb` calculations. I added the dipole gradient to the printout as well, storing it in the intensities, which is somewhat inline with the hessian being stored in the normal modes.

clavigne:
I agree it is pretty hacky. I don't need the feature right now, as I can just compile my PR branch myself, so if you think it's best to refactor instead, I'm open to it.

marcelmbn:
@clavigne Is this feature still relevant? If yes, would you be open to implement it into the current code?

marvinfriede:
@clavigne ping

PR #372, opened 2020-11-02
Fixing some optimization problems for larger structures

MtoLStoN:
With the newly implemented ALPB solvation models, some errors occurred during the optimization of larger structures in water with the Force Field. Implementing a threshold for the descreening parameters in the fitting process for water fixes this error. A similar problem for CH2CL2 was fixed as well.


awvwgk (changes requested):
The GFN-FF SDF test is using the water parameters for ALPB in the second and third test iteration here:

https://github.com/grimme-lab/xtb/blob/0ad987d1257e40814cc1b57920c51139c2f420ea/TESTSUITE/gfnff.f90#L610-L613

Can you update the reference energies and gradient norms there as well?

MtoLStoN:
Sure thing.

awvwgk (approved):
Looks good to me

PR #375, opened 2020-11-06
Fix compilation warnings 

FulgurIgor:
This patch fixes some warnings which were produced when XTB is built.

1. For all iso_fortran_env module imports, `only` statement was added for removing the following warning:
```
Warning: Use of the NUMERIC_STORAGE_SIZE named constant from intrinsic module ISO_FORTRAN_ENV at (1) is incompatible with option -fdefault-integer-8
```
2. Legacy extension with comma before i/o item list was removed (src/gfnff/gfnff_eg.f90):
```
Warning: Legacy Extension: Comma before i/o item list at (1)
```
3. Now, precision is always set from xtb_mctc_accuracy module
4. In intpack.f90, all warnings were fixed; Now, DO statements are compatible with ISO Fortran 2018 standard
5. In david2.f90, HP has the following declaration:
https://github.com/grimme-lab/xtb/blob/3372db870de945726a3a564591f3833c61509c90/src/david2.f90#L231
But, in the code, it was used as 
https://github.com/grimme-lab/xtb/blob/3372db870de945726a3a564591f3833c61509c90/src/david2.f90#L285-L287
so, the incorrect addresses of the array were used.
6. constrpot routine has 5 arguments:
https://github.com/grimme-lab/xtb/blob/3372db870de945726a3a564591f3833c61509c90/src/constr.f90#L158
But in some cases 6 arguments were used:
https://github.com/grimme-lab/xtb/blob/3372db870de945726a3a564591f3833c61509c90/src/constrain_pot.f90#L360
7. Bond type in src/main/property.f90 was fixed: it should be real(8), not integer
8. In tmgrad.f90, all warnings were fixed. It requires additional checks.
9. Now, in help routine Logical Unit can be passed
10. raise routine now not used mode (only two arguments are used)
11. 85 test was fixed for Debug mode. The error was:
```
At line 107 of file /dev/shm/xtb/src/type/topology.f90
Fortran runtime error: Array bound mismatch for dimension 1 of array 'item' (2/3)
```


awvwgk:
Thank you very much for sharing. Most of the changes look fine, I will have a detailed look over the whole PR soon.

src/intpack.f90
```diff
@@ -1413,9 +1410,8 @@ subroutine opaa0(l,m,n,ga,v,d)
       itot=l+m+n
       itoth=itot/2
       pre=(2.d0*pi/ga)*fn(l+1)*fn(m+1)*fn(n+1)
-      if(2*itoth-itot) 1,2,1
-   1  pre=-pre
-   2  del=.25d0/ga
+      pre=-pre
+      del=.25d0/ga
```
awvwgk, line 1414: This part implements a multiplication with `-1**(itot)` using an arithmetic if and integer division: The condition in the arithmetic if should be replaced by `2*itoth-itot` -> `2*(itot/2)-itot` -> `-modulo(itot, 2)`, therefore the correct refactor would be

```f90
if (modulo(itot, 2) == 1) pre = -pre
```

src/intpack.f90
```diff
@@ -983,7 +983,7 @@ pure subroutine opac3(l,m,n,ga,v,d)
    !           electronic part of charge density
    implicit none
    integer,intent(in)  :: l,m,n
-   real(wp), intent(in)  :: ga,d(1)
+   real(wp), intent(in)  :: ga,d(3)
```
awvwgk, line 986: `d(1)` was used as assumed shape array here -> `d(*)`. `d` is indeed a three element vector.

src/intpack.f90
```diff
@@ -1006,7 +1006,7 @@ pure subroutine opad1(l,m,n,ga,v,d)
    !           electronic part of overlap
    implicit none
    integer,intent(in)  :: l,m,n
-   real(wp), intent(in)  :: ga,d(1)
+   real(wp), intent(in)  :: ga,d(3)
```
awvwgk, line 1009: Again, `d(1)` was used to express an assumed size array.

src/intpack.f90
```diff
@@ -1367,16 +1367,13 @@ subroutine fmc(mvar,xvar,expmx,fmch)
...
+         b=b-1.d0
+         approx=approx*b
+      enddo
    endif
-   21  do i=1,m
-      b=b-1.d0
-      approx=approx*b
-   enddo
-   23  fimult=.5d0*expmx*xd
+   fimult=.5d0*expmx*xd
```
awvwgk, line 1376: `goto 23` was used to jump to the else branch of the condition `m != 0`.

src/intpack.f90
```diff
@@ -1476,43 +1472,48 @@ double precision function aainer(l,m,n,d,dp,fnu,fn,fd)
...
-   11     fu=fu*d(1)**(ilru-1)
-   12     last5 = jmsh+1
-          do 2 iv=1,last5
+          if(dabs(d(1))-0.0000001d0.le.0.0) then
+           if(ilru-1.gt.0) cycle
+          else
+           fu=fu*d(1)**(ilru-1)
+          end if
+          last5 = jmsh+1
+          do iv=1,last5
```
awvwgk, line 1486: Second arithmetic if leads to end of loop with label `3` (-> `cycle`). Refactor captures logic correctly.

src/intpack.f90
```diff
@@ -1476,43 +1472,48 @@ double precision function aainer(l,m,n,d,dp,fnu,fn,fd)
...
-   21      fuv=fuv*d(2)**(jmsv-1)
-   22      last6 = knth+1
-           do 1 iw=1,last6
+           if(dabs(d(2))-0.0000001d0.le.0d0) then
+            if(jmsv-1.gt.0) cycle
+           else
+            fuv=fuv*d(2)**(jmsv-1)
+           end if
+           last6 = knth+1
+           do iw=1,last6
```
awvwgk, line 1496: Second arithmetic if leads to end of loop with label `2` (-> `cycle`). Refactor captures logic correctly.

src/intpack.f90
```diff
@@ -1476,43 +1472,48 @@ double precision function aainer(l,m,n,d,dp,fnu,fn,fd)
...
-            if(dabs(d(3))-0.0000001d0)30,30,31
-   30       if(kntw-1) 32,32,1
-   31       fuvw=fuvw*d(3)**(kntw-1)
-   32       uvwt=u+v+w
+            if(dabs(d(3))-0.0000001d0.le.0d0) then
+             if(kntw-1.gt.0) cycle
+            else
+             fuvw=fuvw*d(3)**(kntw-1)
+            end if
+            uvwt=u+v+w
```
awvwgk, line 1505: Second arithmetic if leads to end of loop with label `1` (-> `cycle`). Refactor captures logic correctly.

src/intpack.f90
```diff
@@ -1476,43 +1472,48 @@ double precision function aainer(l,m,n,d,dp,fnu,fn,fd)
...
             uvwth=uvwt/2
-            if(2*uvwth-uvwt) 33,34,33
-   33       fuvw=-fuvw
-   34       nuindx=lmrsij-uvwt
+            fuvw=-fuvw
+            nuindx=lmrsij-uvwt
```
awvwgk, line 1508: Arithmetic if in combination with integer arithmetic implements `-1**(uvwt)`.
Condition of arithmetic if should be refactored as: `2*uvwth-uvwt` -> `2*(uvwt/2)-uvwt` -> `-modulo(uvwt, 2)`.

Therefore, the refactored code should read:
```f90
if (modulo(uvwt, 2) == 1) fuvw = -fuvw
```

src/set_module.f90
```diff
@@ -514,6 +512,8 @@ subroutine write_set_constrain(ictrl)
          write(ictrl,'(3x,"dihedral:",1x,4(i0,",",1x),g0,1x,"# in Degree")') &
             atconstr(1,i),atconstr(2,i),atconstr(3,i),atconstr(4,i), &
             valconstr(i)*180.0_wp/pi
+      case default ! this is bad, we don't want this to happen
+         call raise('E','This is an internal error, please report this!')
```
awvwgk, line 516: Error handling should happen at the top of the select case, for readability reasons. Could you please revert the changes in this source file?

src/esp.f
```diff
@@ -383,11 +382,12 @@ subroutine propa1(aname,c,nnn,gama,e,dd,va)
...
-   41 call aname(lin(i),min(i),nin(i),gama,v,d)
-      va=dd(i)*v+va
-   61 continue
+      do i=1,nnn
+         if (dabs(dd(i))-1.d-8 .gt. 0.d0) then
+            call aname(lin(i),min(i),nin(i),gama,v,d)
+            va=dd(i)*v+va
+         end if
+      end do
       return
```
awvwgk, line 391: Arithmetic if used to implement `cycle`. Refactored correctly.

src/david2.f90
```diff
@@ -283,7 +283,7 @@ subroutine solver_sdavidson(n,crite,Hp,C,e,fail,pr)
    e    = 0
 
    do i=1,n
-      adiag(i)=HP(i*(i-1)/2)
+      adiag(i)=HP(i*(i+1)/2)
```
awvwgk, line 286: Diagonal element is given by `i*(i-1)/2+i` == `i*(i+1)/2`. Fixes out-of-bound access.

src/constrain_pot.f90
```diff
@@ -357,10 +357,10 @@ subroutine constrain_hess(fix,n,at,xyz0,Hess)
          ii = (ia-1)*3+ic
          xyz(ic,ia)=xyz0(ic,ia)+step
          gr=0.0_wp
-         call constrpot(fix,n,at,xyz,gr,e)
+         call constrpot(n,at,xyz,gr,e)
          xyz(ic,ia)=xyz0(ic,ia)-step
          gl=0.0_wp
-         call constrpot(fix,n,at,xyz,gl,e)
+         call constrpot(n,at,xyz,gl,e)
```
awvwgk, line 363: This part probably contains dead code, nevertheless, with this refactor the procedure call uses the correct interface.

src/tmgrad.f90
```diff
@@ -38,22 +38,12 @@ subroutine tmgrad(nat,iat,xyz,g,edum)
...
-      a='$end'
-      write(jen,'(a)')a
-      goto 22
-   endif
-   goto 20
-22 continue
+   read(ien,'(a)')a
+   write(jen,'(i6,3F18.11)')j+1,edum,edum,edum
+   a='$end'
+   write(jen,'(a)')a
```
awvwgk, line 46: This refactoring is incorrect, the energy file usually looks like:
```
$energy
 1  <real> <real> <real>
 2  <real> <real> <real>
...
$end
```

The routine reads the file complete file and inserts a new line when the `$end` group is encountered.

src/tmgrad.f90
```diff
@@ -66,30 +56,20 @@ subroutine tmgrad(nat,iat,xyz,g,edum)
...
+   &           ''|dE/dxyz| ='',F10.6)')j+1,edum,gsum
+   do i=1,nat
+      write(jgr,'(3(F20.14,2x),4x,a2)')xyz(1,i),xyz(2,i),xyz(3,i), &
+      &                                toLcSymbol(iat(i))
+   enddo
+   do i=1,nat
+      write(jgr,'(3D22.13)')g(1,i),g(2,i),g(3,i)
+   enddo
+   a='$end'
+   write(jgr,'(a)')a
```
awvwgk, line 72: Similar to above, this refactor is incorrect. The gradient file is read in and a new geometry + gradient inserted before the `$end` group.

awvwgk (commented):
I'm fine with most changes, except for the moved default case in `src/set_module.f90`, please revert this change.

The code in `intpack.f90` is clearly legacy code, the part modified is only used for some special run modes and therefore not tested with the usual testsuite. I'm a bit torn here, while I would certainly like to purge all obsolete and deleted Fortran features from the code base, there is quite a risk associated with refactoring this piece.

I tried to comment on every change in `intpack.f90` to describe the intent of the original implementation and spotted two incorrect refactoring changes there. This might make this refactoring saver.

src/intpack.f90
FulgurIgor: Should I convert it to `d(*)`, or my changes accepted? 
I changed it for fixing the following warning:
```
/dev/shm/xtb/src/intpack.f90:992:15:

  991 |    do i=1,3
      |           2    
  992 |       dd(i)=-d(i)
      |               1
Warning: Array reference at (1) out of bounds (3 > 1) in loop beginning at (2)
```

src/intpack.f90
FulgurIgor: Could you please expand on what do you mean? For me, changes look correct.

src/intpack.f90
awvwgk: It is fine that way, I was just commenting on the original intent such that we can compare notes here.

src/intpack.f90
awvwgk: The change is correct, I was just taking notes while reviewing.

src/tmgrad.f90
awvwgk: The gradient file is given as:
```
$grad
  cycle =       1    SCF energy = <real> |dE/dxyz| = <real>
  <x1> <y1> <z1> <sym1>
  <x2> <y2> <z2> <sym2>
  ...
  <gx1> <gy1> <gz1>
  <gx2> <gy2> <gz2>
  ...
  cycle =       2    SCF energy = <real> |dE/dxyz| = <real>
  <x1> <y1> <z1> <sym1>
  <x2> <y2> <z2> <sym2>
  ...
  <gx1> <gy1> <gz1>
  <gx2> <gy2> <gz2>
  ...
$end
```

A new gradient entry is append together with the corresponding coordinates when the `$end` group is encountered.

I have a refactored implementation in [`dftd4`](https://github.com/dftd4/dftd4/blob/9762d50418c0d9859f24005e4e63631b75d85792/source/disp_output.f90#L124-L204) and [`s-dftd3`](https://github.com/awvwgk/simple-dftd3/blob/0fbd084df9b6b140f20362fa832d324ea792e32b/src/dftd3/output.f90#L350-L427).

FulgurIgor: I have looked through the code to find places where tmgrad routine is used, but I did not find any. Can I remove this file?

The output of grep is the following:
```$ grep "tmgrad" * -irn
src/tmgrad.f90:18:subroutine tmgrad(nat,iat,xyz,g,edum)
src/tmgrad.f90:97:end subroutine tmgrad
src/meson.build:147:  'tmgrad.f90',
src/CMakeLists.txt:149:  "${dir}/tmgrad.f90"
```

awvwgk: Oh, it is dead code, feel free to remove it than.

awvwgk (approved):
Thank you very much.

PR #385, opened 2020-11-24
Fix end of lines (CLRF -> CL)

FulgurIgor:
This patch removes all Windows-style end of lines (CLRF) in all files. The file .gitattributes was added for the default end of lines in CL style. Also, tabs were removed in source files.

The problem with CLRF was noticed when I edited david.f file in mc.


awvwgk (commented):
A correctly configured editor can easily cope with the fact that some source files use Windows style line terminator and some use Unix style line terminators.

I don't see any advantages of introducing a huge diff for just a whitespace change.

PR #387, opened 2020-11-27
Reoptimized Water, added Methanol and Ethanol

MtoLStoN:
Fixed problems with charged Compounds containing phosphor for GFN2/ALPB by reoptimizing the water parameters.

Parametrized the solvents Methanol and Ethanol for the ALPB Method.


awvwgk (approved):
Thanks.

PR #388, opened 2020-11-29
Sign error fixed for unbiased frequencies in SPH calculations

sespic:
Fixed inverse sign of frequencies for unbiased SPH calculations


awvwgk (approved):
Looks good to me.

PR #399, opened 2020-12-18
Fixing IPEA-xTB

haneug:
This simple change fixes IPEA-xTB. I only tested it sporadically so I can't ensure if this does not break anything else. Maybe we wait a little while with merging as I keep testing this change.


awvwgk (commented):
I created a unit test with your indole example from the issue. In case you build and run the test *outside* of the testsuite:
```
./build_oneapi/TESTSUITE/xtb_test gfn1 ipea
...
normal termination of test
```
Everything works fine, but if you run it *inside* the testsuite it will fail due to `_MALLOC_PERTURB` with:
```
libifcore.so.5     00007FF95ECADD29  for__signal_handl     Unknown  Unknown
libpthread-2.32.s  00007FF95E1B00F0  Unknown               Unknown  Unknown
xtb_test           00000000007C006D  xtb_type_wavefunc          58  wavefunction.f90
xtb_test           0000000000737959  xtb_main_setup_mp         293  setup.f90
xtb_test           0000000000516F5E  test_ipea_indole_         796  gfn1.f90
xtb_test           0000000000581967  MAIN__                     93  tests_peeq.f90
xtb_test           0000000000408122  Unknown               Unknown  Unknown
libc-2.32.so       00007FF95DFFB152  __libc_start_main     Unknown  Unknown
xtb_test           000000000040802E  Unknown               Unknown  Unknown
```

haneug:
Thank you for creating this test. I think that the reason for the segmentation fault could be that the parameter file 'param_ipea-xtb.txt' is present when you call the test outside the testsuite, but inside it is missing and you have no parameters. I will look into this.

awvwgk:
You are right, we have to search for the parameter file rather than relying for the default here.

awvwgk:
This should take care of the parameter file issue.

awvwgk (approved):

PR #407, opened 2020-12-31
Support NVHPC compilers in meson build

awvwgk:
Add tests for NVHPC signature and treat them similar to PGI compilers.


meson/meson.build
```diff
@@ -32,7 +32,7 @@ elif fc.get_id() == 'intel'
   add_project_arguments('-axAVX2',    language: 'fortran')
   add_project_arguments('-r8',        language: 'fortran')
   add_project_arguments('-traceback', language: 'fortran')
-elif fc.get_id() == 'pgi'
+elif fc.get_id() == 'pgi' or fc.get_id() == 'nvidida_hpc'
```
awvwgk, line 35:
```suggestion
elif fc.get_id() == 'pgi' or fc.get_id() == 'nvidia_hpc'
```

PR #416, opened 2021-01-16
Implement support for ELPA solver

awvwgk:
This PR implements support for the ELPA solver (onenode and onenode_openmp variant). Tested with GCC toolchain and ELPA 2020.11.001. No successful build with Intel toolchain so far.

For small systems (taxol, 113 atoms @ 2 core Sky Lake) the ELPA solver can be by a factor two slower for than the MKL solver.

The main issue results from a portion of ELPA which is running always sequentially on a single process with multiple threads. This breaks the efficiency for multi-threaded applications. Maybe it is possible to select a threaded algorithm instead and get a fully threaded ELPA run.

Needs further investigation.

### DIFF
--- a/meson/meson.build
+++ b/meson/meson.build
@@ -147,6 +147,24 @@ if get_option('openmp')
   dependencies += omp_dep
 endif
 
+if get_option('elpa') != 'none'
+  if get_option('openmp')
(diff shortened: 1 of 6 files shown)


awvwgk:
The shared memory (OpenMP) version of ELPA was found to be unfeasible due to significantly larger runtimes compares to Intel's MKL.

PR #445, opened 2021-02-19
Update References

awvwgk:
- WIREs review is now printed


src/xhelp.f90
```diff
@@ -58,6 +58,10 @@ subroutine citation(iunit)
    "for metadynamics refer to:",&
    "* S. Grimme, J. Chem. Theory Comput., 2019, 155, 2847-2862", &
    "  DOI: 10.1021/acs.jctc.9b00143", &
+   "",&
+      "for SPH calculations refer to:",&
```
awvwgk, line 62:
```suggestion
   "for SPH calculations refer to:",&
```

Issue #458, opened 2021-03-18
hess calculation seg fault


ucb-mgcf:
**Describe the bug**
I'm getting a SIGSEGV seg fault for a --hess calculation.  The structure was successfully optimized in xtb.

**To Reproduce**
Steps to reproduce the behaviour:

1. happens with input: 
[start.txt](https://github.com/grimme-lab/xtb/files/6167717/start.txt)

2. start `xtb` with xtb start.xyz --ohess extreme > output.  Although I've also had the same result when using xtb start.xyz --chrg 0 --uhf 0 --hess --grad > xtbout
3. output is same for --verbose 
4. last printout before seg fault is for the time estimates for hessian calculation

Please provide all input and output file such that we confirm your report.

**Expected behaviour**
hess calculation should run to completion

**Additional context**
I'm using binary release, xtb version 6.4.0 (d4b70c2).  I saw a couple of earlier related bugs from last year, but these were fixed before this version.

awvwgk:
Can confirm that it segfaults with 6.4.0 and latest head (9c8c9fdf7fc760ef3a1a1161954ed8db39bbc47e), not sure what is causing it but it seems to be related to the parallelization on the displacements in the hessian. Inverting the distribution of the cores in the hessian `OMP_NUM_THREADS=1,32` instead of `OMP_NUM_THREADS=32,1` runs without problems, putting in more than one core on the hessian parallelization triggers the bug.

ucb-mgcf:
Thanks for the reply.  I've tried this fix and it works for me.  Would you suggest using this number-of-threads inversion generally, or only when this error occurs?

awvwgk:
Not really a fix but more a diagnostic here, using the threads to parallelize over the displacements will be more efficient than using them to in the electronic structure method while evaluating the displacements sequentially.

There are several issues with the numerical differentiation I have to look into, will do so soon.

PR #465, opened 2021-04-17
Fix parallelization strategy for numerical differention

awvwgk:
- Intel Fortran has problems to properly handle orphaned OpenMP

@ucb-mgcf, @sespic, @fabothch Do you mind rechecking your failing examples with the version uploaded at:
https://github.com/awvwgk/xtb/suites/2519893098/artifacts/54567679

Closes https://github.com/grimme-lab/xtb/issues/458

### DIFF
--- a/TESTSUITE/hessian.f90
+++ b/TESTSUITE/hessian.f90
@@ -127,7 +127,7 @@ subroutine test_gfn1_hessian
    dipgrad = 0.0_wp
    hessian = 0.0_wp
    ! omp parallel default(shared)
-   call numdiff2(env, mol, chk, calc, step, hessian, dipgrad)
+   call numdiff2(env, mol, chk, calc, step, hessian, dipgrad, .false.)
    ! omp end parallel
 
    do i = 1, size(dipgrad_ref, 2)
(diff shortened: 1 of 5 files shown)


fabothch:
The system I investigated works now, when using GFN-FF.  From my side this appears to work.

ucb-mgcf:
The above link to the version for me to test appears to be broken.  Should I test this by installing the latest master branch?

awvwgk:
GitHub invalidates the uploaded artifacts quite quickly, the patch is included in the version on the bleed tag as well now:
https://github.com/grimme-lab/xtb/releases/tag/bleed

ucb-mgcf:
I just confirmed that this calculation now works using the original specification, i.e. OMP_NUM_THREADS=16,1 with the bleeding-edge version.  Thanks!  For efficiency, would you now recommend continuing to use OMP_NUM_THREADS=1,n or go back to using OMP_NUM_THREADS=n,1?

awvwgk:
n,1 distribution should work best, the more easy to exploit parallelism is in the displacements of the hessian instead of the individual SCC calculations. You can try to set `OMP_SCHEDULE=dynamic` as well, in some cases it can improve the load balancing on the threads.

PR #470, opened 2021-04-21
Implementation of the analytical TURBOMOLE Hessian

cplett:
The TURBOMOLE interface was extended by the analyitcal hessian computation. It will be performed by default, if the flag --tm is used in combination with --bhess, --ohess or --hess.

### DIFF
--- a/src/freq/meson.build
+++ b/src/freq/meson.build
@@ -20,4 +20,5 @@ srcs += files(
   'numdiff.f90',
   'project.f90',
   'utils.f90',
+  'turbomole.f90'
 )
--- a/src/freq/turbomole.f90
+++ b/src/freq/turbomole.f90
@@ -52,7 +52,6 @@ subroutine aoforce_hessian(env, mol, h, dipd)
    call reader%open('hessian')
    call readHessian(env, mol, h, reader, fileType%tmol)
    call reader%close
-   call writeHessianOut('hessian',h)
 
    call open_file(idipd,'dipgrad','r')
    if(idipd == -1) then
@@ -69,33 +68,33 @@ subroutine aoforce_hessian(env, mol, h, dipd)
 
    call close_file(idipd)
 
-   end subroutine aoforce_hessian
+end subroutine aoforce_hessian
 
 
-   subroutine read_dipgrad(idipd, n, dipd, error)
-        use xtb_mctc_systools
+subroutine read_dipgrad(idipd, n, dipd, error)
+   use xtb_mctc_systools
 
-        implicit none
+   implicit none
 
-        integer, intent(out) :: error
-        character(len=:), allocatable :: line
-        integer, intent(in) :: n
-        real(wp), intent(inout) :: dipd(:,:)
-        integer, intent(in) :: idipd
-        integer :: i,j
+   integer, intent(out) :: error
+   character(len=:), allocatable :: line
+   integer, intent(in) :: n
+   real(wp), intent(inout) :: dipd(:,:)
+   integer, intent(in) :: idipd
+   integer :: i,j
 
-        error = 0
+   error = 0
 
-        do while(error == 0)
-                call getline(idipd, line, error)
-                if (index(line, '$dipgrad          cartesian dipole gradients') == 1) then
-                        do i=1, 3*n
-                                read(idipd,*) (dipd(j,i),j=1,3)
-                        end do
-                        exit 
-                end if
-        enddo
-   end subroutine read_dipgrad
+   do while(error == 0)
+      call getline(idipd, line, error)
+         if (index(line, '$dipgrad          cartesian dipole gradients') == 1) then
+            do i=1, 3*n
+               read(idipd,*) (dipd(j,i),j=1,3)
+            end do
+            exit 
+         end if
+   enddo
+end subroutine read_dipgrad
  
 
 end module xtb_freq_turbomole 
--- a/src/freq/turbomole.f90
+++ b/src/freq/turbomole.f90
@@ -1,0 +1,101 @@
+! This file is part of xtb.
+!
+! Copyright (C) 2021 Christoph Plett
+!
+! xtb is free software: you can redistribute it and/or modify it under
+! the terms of the GNU Lesser General Public License as published by
+! the Free Software Foundation, either version 3 of the License, or
+! (at your option) any later version.
+!
+! xtb is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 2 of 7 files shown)


src/prog/main.F90
```diff
@@ -756,8 +756,12 @@ subroutine xtbMain(env, argParser)
    ! ------------------------------------------------------------------------
    !> numerical hessian calculation
    if ((runtyp.eq.p_run_hess).or.(runtyp.eq.p_run_ohess).or.(runtyp.eq.p_run_bhess)) then
-      if (runtyp.eq.p_run_bhess) then
+      if (runtyp.eq.p_run_bhess .and. mode_extrun.ne.p_ext_turbomole) then
          call generic_header(env%unit,"Biased Numerical Hessian",49,10)
+      else if (runtyp.eq.p_run_bhess .and. mode_extrun.eq.p_ext_turbomole) then
+         call generic_header(env%unit,"Biased Analytical Hessian",49,10)
+      else if (mode_extrun.eq.p_ext_turbomole) then
+         call analyTMhess_header(env%unit)
```
awvwgk, line 764: Can this be done via a `generic_header` instead?

src/hessian.F90
```diff
@@ -393,7 +409,7 @@ subroutine numhess( &
    if(res%linear)then
       write(env%unit,'(1x,a)') 'vibrational frequencies (cm⁻¹)'
    else
-      write(env%unit,'(1x,a)') 'projected vibrational frequencies (cm⁻¹)'
+      write(env%unit,'(1x,a)') 'Eprojected vibrational frequencies (cm⁻¹)'
```
awvwgk, line 412: Is this change meant to be committed?

src/freq/turbomole.f90
```diff
@@ -0,0 +1,101 @@
...
+           call env%error('An error occurred while reading the dipolegradient', source)
+           return
+   end if
+
+   call close_file(idipd)
+
+   end subroutine aoforce_hessian
+
+
+   subroutine read_dipgrad(idipd, n, dipd, error)
```
awvwgk, line 75: The indentation here is off, we are generally using three spaces indentation inside, procedure bodies or if / do blocks.

src/freq/turbomole.f90
```diff
@@ -0,0 +1,101 @@
...
+   real(wp), intent(inout) :: h(:,:)
+   real(wp), intent(inout) :: dipd(:,:)
+   integer :: idipd, error
+
+   call wrtm(mol%n,mol%at,mol%xyz) !Overwrite coord with RAM-xyz file
+   call execute_command_line('exec aoforce > job.last2>> /dev/null')
+   call reader%open('hessian')
+   call readHessian(env, mol, h, reader, fileType%tmol)
+   call reader%close
+   call writeHessianOut('hessian',h)
```
awvwgk, line 55: Is there a reason to write the hessian here directly after reading it?

src/hessian.F90
```diff
@@ -175,6 +176,13 @@ subroutine numhess( &
 !! ========================================================================
 !  Hessian part -----------------------------------------------------------
 
+   !analytical hessian calculation
+   if(mode_extrun .eq. p_ext_turbomole) then    
+           dipd=0.0_wp
+           freezeset%n=0
```
awvwgk, line 182: Overwriting an input variable is generally a bad idea, if we can't support this runmode we should fail rather than doing something else here.

src/hessian.F90
```diff
@@ -175,6 +176,13 @@ subroutine numhess( &
...
+           call aoforce_hessian(env,mol,h,dipd) 
+           call writeHessianOut('hessian',h)
```
awvwgk, line 184: Is there a reason to write the hessian here?

src/hessian.F90
```diff
@@ -234,11 +242,19 @@ subroutine numhess( &
...
+   endif !From turbomole_exception
+
+
 !  Hessian done -----------------------------------------------------------
 !! ========================================================================
 
    if (runtyp.eq.p_run_bhess) call numhess_rmsd(env,mol,hbias)
 
+   if(mode_extrun .eq. p_ext_turbomole .AND. runtyp.eq.p_run_bhess) then 
+        htb = h !htb equals unbiased hessian
```
awvwgk, line 254: Is this assignment here required, `htb` will be overwritten by assignment later.

src/prog/main.F90
```diff
@@ -1047,8 +1051,10 @@ subroutine xtbMain(env, argParser)
    if (runtyp.eq.p_run_path) then
       call prtiming(4,'path finder')
    endif
-   if ((runtyp.eq.p_run_hess).or.(runtyp.eq.p_run_ohess).or.(runtyp.eq.p_run_bhess)) then
+   if (((runtyp.eq.p_run_hess).or.(runtyp.eq.p_run_ohess).or.(runtyp.eq.p_run_bhess)) .and. (mode_extrun.ne.p_ext_turbomole)) then
       call prtiming(5,'numerical hessian')
+   else if (((runtyp.eq.p_run_hess).or.(runtyp.eq.p_run_ohess).or.(runtyp.eq.p_run_bhess)) .and. (mode_extrun.eq.p_ext_turbomole)) then
+      call prtiming(5,'analytical hessian')
```
awvwgk, line 1057: Let's simplify this logic a bit and move the condition for the timer label inside the if block.
```suggestion
   if (((runtyp.eq.p_run_hess).or.(runtyp.eq.p_run_ohess).or.(runtyp.eq.p_run_bhess))) then
      if (mode_extrun.ne.p_ext_turbomole) then
         call prtiming(5,'analytical hessian')
      else
         call prtiming(5,'numerical hessian')
      end if
```
Removing the numerical / analytical attribute would also work.

src/freq/meson.build
```diff
@@ -20,4 +20,5 @@ srcs += files(
   'numdiff.f90',
   'project.f90',
   'utils.f90',
+  'turbomole.f90'
```
awvwgk, line 23: This needs to be added to the CMakeLists.txt as well

awvwgk (changes requested):
Thanks for sharing, Christoph. I left a few comments. Let me know if you need any help from Sebastian @sespic or me.

src/freq/turbomole.f90
cplett: It will overwrite the hessian file in turbomole format with a hessian file in the xtb format. If this is not required, I will change it.

awvwgk: This might turn out produce interesting results in case you run a namespaced calculation (`--namespace <prefix>`).

cplett: I've got your point. It will be removed. Thank you.

cplett:
Thank you @awvwgk for your comments. I included the suggestions.

PR #481, opened 2021-05-17
Multipole experiments

awvwgk:
- [x] Correct sign of multipolar potential terms (sign from population should be accounted for in H1 rather than the potential)
- [ ] Origin independence of the integrals
- [ ] Remove origin dependent terms from potential
- [ ] Trace removal should be part of the integral and not of the population
- [ ] Ordering of multipole moments should be consistent in integral, potential and populations


awvwgk:
Closing in favor of interfacing `xtb` with `tblite`.

Issue #451, opened 2021-03-05
Detection of SCF non-convergence via C-API


steinmig:
**Is your feature request related to a problem? Please describe.**
I could not find a way to detect a non-converged SCF via the C-API.

**Describe the solution you'd like**
Get a proper error code for a non-converged SCF

**Describe alternatives you've considered**
In case this is against your design philosophy since the program ended normally, I tried two other things:

1) I tried the `xtb_getError()` call to get some std err information, but that was empty although the SCF did not converge. I cannot outrule that I made a mistake here, but I figured that you do not write the unconverged SCF into the error out, since your program is also continuing normally.

2) I then went around both of that and tried to parse the output of xtb. Here, I encountered the problem that I could not manage to release the output. I was able to call `xtb_setOutput()` and `xtb_releaseOutput()`. However, I did not get any error for the release, but still could not access the content of the outputfile (I only managed to get the very first line). Even after proper:
```
    xtb_delResults(&res);
    xtb_delCalculator(&calc);
    xtb_delMolecule(&mol);
    xtb_delEnvironment(&env);
```
I could not access the xtb output file. Only after my script itself was finished the complete output file was accessible in the console.

So the three solutions I see:
 - Change exit code (faster and easier for anything wrapping xtb, but maybe against your guidelines)
 - Write unconverged SCF into the stderr
 - Fixing the release of the output file (in case I did not make any mistakes)

**If possible state how you can assist in providing data or code to to implement the feature**
Since my wrapper is not released yet and has some more dependencies, it's hard for me to provide you with a MWE. I can try to set something up, in case it is necessary.

awvwgk:
A non-convergend SCF is a particular case, I can see good reasons to raise it as an error, but there are also cases where an initially non-converged SCF can still be used, even if they are rare.

The easiest way to resolve this:
1. make failing SCF convergence an error rather than just warn
2. allow the `xtb` main driver to rescue this error and continue
3. propagate this error to the C-API with a dedicated error code to allow proper handling

steinmig:
@awvwgk Thanks for the easy solution for the API user!

Do you have any update for me concerning a time estimate when this will be implemented?

awvwgk:
The relevant logic is here:
https://github.com/grimme-lab/xtb/blob/f63f532eb023642304911129b8a6c99c78d82aaf/src/scf_module.F90#L645-L648

[...]

PR #485, opened 2021-05-21
Change unconverged SCF to error

steinmig:
Closes https://github.com/grimme-lab/xtb/issues/451 based on suggestion of @awvwgk

### DIFF
--- a/src/scf_module.F90
+++ b/src/scf_module.F90
@@ -634,18 +634,16 @@ subroutine scf(env, mol, wfn, basis, pcem, xtbData, solvation, &
 
    !$acc exit data delete(s, dpint, qpint)
 
-   ! check if something terrible happend in the SCC
+   ! check if something terrible happened in the SCC
    call env%check(exitRun)
    if (exitRun) then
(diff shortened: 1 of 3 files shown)


src/scf_module.F90
```diff
@@ -661,6 +659,11 @@ subroutine scf(env, mol, wfn, basis, pcem, xtbData, solvation, &
    if (.not.pr.and.profile.and.minpr) &
       call timer%write_timing(env%unit,5,'SCC iter.')
    if (profile) call timer%measure(6,"molecular gradient")
+   
+   ! throw error for unconverged SCF
+   if (fail) then
+      call env%error("Self consistent charge iterator did not converge", source)
+   end if
```
awvwgk, line 666: Let's move this error after the gradient calculation and before the printing. Some calculations access the environment and might abort the calculation if an error is present.

awvwgk (approved):
Thanks, looks good now. I'll mark it for inclusion in version 6.4.1

steinmig:
@awvwgk Great thanks for the review!

PR #497, opened 2021-06-29
Writing adjacency to output

hoelzerC:
Optional output of adjacency and coordination number for GFN-FF 

Signed-off-by: C.Hoelzer <hoelzer@thch.uni-bonn.de>

### DIFF
--- a/src/gfnff/gfnff_setup.f90
+++ b/src/gfnff/gfnff_setup.f90
@@ -230,12 +230,9 @@ subroutine write_gfnff_adjacency(fname, topo)
    call open_file(ifile,fname,'w') 
    ! looping over topology neighboring list
    if (ifile.ne.-1) then
-      write(ifile, *) '# indices of neighbouring atoms (max seven); connectivity given by last entry;'  
-      do i = 1, size(topo%nb,2) ! #atoms
-         do j = 1, size(topo%nb,1) ! 20
-            write(ifile, '(I2,X)', advance='no') topo%nb(j,i)
-         end do
-         write(ifile, *) ''  ! line break
+      write(ifile, '(a)') '# indices of neighbouring atoms (max seven)'  
+      do i = 1, size(topo%nb, 2)
+        write(ifile, '(*(i0:, 1x))') (topo%nb(j, i), j = 1, topo%nb(size(topo%nb, 1), i))
       end do
    end if
    call close_file(ifile)
--- a/src/gfnff/gfnff_setup.f90
+++ b/src/gfnff/gfnff_setup.f90
@@ -87,6 +87,7 @@ subroutine gfnff_setup(env,verbose,restart,mol,gen,param,topo,accuracy,version)
 
   if (.not.mol%struc%two_dimensional) then
      call write_restart_gff(env,'gfnff_topo',mol%n,version,topo)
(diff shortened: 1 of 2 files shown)


awvwgk (commented):
The neighbour/adjacency list is not really an output of the calculation, therefore I don't think it makes sense to write it with the singlepoint calculation. Preferably, we can generate such output with the topology generation or extract it from the topology restart file directly.

One possibility would be to create such output in the topology subcommand when generating a topology restart file ahead of an actual singlepoint calculation:
https://github.com/grimme-lab/xtb/blob/master/src/prog/topology.f90

hoelzerC:
The method was refactored into the gfnff_setup routine. Now, the adjacency list is written with the topo restart file (write_restart_gff).

src/gfnff/gfnff_setup.f90
```diff
@@ -218,4 +219,28 @@ subroutine gfnff_input(env, mol, topo)
...
+   type(TGFFTopology) :: topo
+   integer :: i, j
+
+   call open_file(ifile,fname,'w') 
+   ! looping over topology neighboring list
+   if (ifile.ne.-1) then
+      write(ifile, *) '# indices of neighbouring atoms (max seven); connectivity given by last entry;'  
+      do i = 1, size(topo%nb,2) ! #atoms
+         do j = 1, size(topo%nb,1) ! 20
+            write(ifile, '(I2,X)', advance='no') topo%nb(j,i)
```
awvwgk, line 236: This statement has several issues, first since you are using a non-advancing output it will produce a trailing whitespace from the format specifier. Second, leaving away the width of the `x` identifier is a standard extension and we want to avoid relying on those.

Third, the integer field will overflow for more than 99 atoms:

```
❯ tail gfnff_adjacency
** ** **  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  3  
** ** **  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  3  
** ** **  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  3  
** ** **  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  3  
**  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  
**  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  
**  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  
**  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  
**  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1  
86  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  0  1
```

Finally, printing the padded adjacency list exposes an implementation detail like the preallocated number of entries for each atom, which is fragile in case we change this internally. Maybe a more practical format might be

```f90
do i = 1, size(topo%nb, 2)
  write(ifile, '(*(i0:, 1x))') topo%nb(20, i), (topo%nb(j, i), j = 1, topo%nb(20, i))
end do
```

hoelzerC: Good idea, I adapted the output format.

PR #538, opened 2021-12-01
Add topology list printout to GFN-FF

Thomas3R:
You can now print GFN-FF topology lists (e.g. nb, bpair, alist, ...) with the --wrtopo flag.

Signed-off-by: Thomas Rose <39367840+Thomas3R@users.noreply.github.com>

### DIFF
--- a/src/main/property.F90
+++ b/src/main/property.F90
@@ -602,8 +602,13 @@ subroutine print_wiberg(iunit,n,at,sym,wbo,thr)
          if (wbr(i,j).gt.thr) ibmax=j
          xsum=xsum+wbr(i,j)
       enddo
-      write(iunit,'(i6,1x,i3,1x,a4,f6.3,1x,"--")',advance='no') &
-         & i,at(i),sym(i),xsum
+      if (ibmax > 0) then
+         write(iunit,'(i6,1x,i3,1x,a4,f6.3,1x,"--")',advance='no') &
+            & i,at(i),sym(i),xsum
+      else
+         write(iunit,'(i6,1x,i3,1x,a4,f6.3)') &
+            & i,at(i),sym(i),xsum
+      end if
       do j = 1, ibmax, 3
          if (j > 1) then
             write(iunit,'(t25)', advance='no')


src/gfnff/gfnff_setup.f90
```diff
@@ -43,6 +43,7 @@ subroutine gfnff_setup(env,verbose,restart,mol,gen,param,topo,accuracy,version)
   logical,intent(in) :: restart
   logical,intent(in) :: verbose
   real(wp),intent(in) :: accuracy
+  integer, intent(in) :: iwrtopo(21)
```
awvwgk, line 46: Let's make this a derived type

src/gfnff/gfnff_setup.f90
```diff
@@ -239,5 +244,116 @@ subroutine write_gfnff_adjacency(fname, topo)
...
+  do i=1, 21
+    if(iwrtopo(i).ne.0) nprint=nprint+1
+  enddo
+
+  do i=1, nprint
+  if(iwrtopo(i).eq.1)then ! nb(20,n) 
+     INQUIRE(FILE='topo_nb', EXIST=f_ex)
+     if(f_ex)then
+       open(43,file='topo_nb')
+       close(43,status='delete')
```
awvwgk, line 268: Let's use our `open_file` and `close_file` routines for this purpose. Also, we don't use hardcoded file units in `xtb`.

src/gfnff/gfnff_setup.f90
```diff
@@ -239,5 +244,116 @@ subroutine write_gfnff_adjacency(fname, topo)
...
+  integer :: n ! num atoms
+  !!! list to print
+  ! num lists to print
+  integer :: nprint
+  integer :: i,j,k
+  integer :: lin
+  logical :: f_ex
+  nprint=0
+  do i=1, 21
+    if(iwrtopo(i).ne.0) nprint=nprint+1
```
awvwgk, line 260: What is that for? Why do we have to loop over nprint here?

awvwgk (changes requested):
This needs some work.

1. please use a derived type to store this data, this is much more self documentating than an array of integers

2. creating another collection of files from `xtb` is not an option. Let's target a standardized format for printing this quantities, like json, npz or similar.

3. if `xtb` can write those files, it might be useful to be able to read them again and test whether we can actually make a round trip with the topology data without losing information.

PR #609, opened 2022-03-18
Fix .CHRG handling in GFN-FF

MtoLStoN:
Handling of the .CHRG file in GFN-FF was broken, if no fragment charges were given. This should fix this.


awvwgk (approved):
Nice catch. Thanks for fixing.

PR #644, opened 2022-06-10
Added rr-ho interpolation for heat capacity in thermo.f90

pprcht:
The RR-HO interpolation is performed for *S*<sub>vib</sub> (<https://doi.org/10.1002/chem.201200497>).
As per communication with S.G., the same type of interpolation *should* also be conducted for *C*<sub>p</sub>.
This implementation does that. Only a very minor change that doesn't affect anything else.
For `--sthr 0` the "original" *C*<sub>p</sub> without interpolation is correctly recovered.


awvwgk (approved):
Thanks for sharing.

PR #648, opened 2022-06-15
Make c-api example MSVC compatible

TyBalduf:
MSVC cl has stricter rules about constants than other compilers
(e.g variables declared `const` are not constant at compile time and so
aren't consider constant). This means the compiler won't allow `const`
variables to be used to declare array sizes.

The changes here explicitly allocate/free memory for arrays so that
the MSVC cl compiler can run this example. The c-api itself should work
without these changes, but the example was incompatible and was halting
compilation.


awvwgk (approved):

PR #654, opened 2022-06-24
Turn off GFN-FF fragmentation if it is not needed.

MtoLStoN:

### DIFF
--- a/src/gfnff/frag_hess.f90
+++ b/src/gfnff/frag_hess.f90
@@ -307,7 +307,7 @@ module xtb_gfnff_fraghess
            ass = maxloc(nspinsyst, 1)
 
             if (nsystem .ge. nspin) then
-               call env%warning("Created more Fragments than atoms. This does not look good. Turning off fragmentization.")
+               call env%warning("Created more fragments than atoms. This does not look good. Turning off fragmentation.")
                nsystem=1
                return
             end if


src/gfnff/gfnff_ini.f90
```diff
@@ -1833,12 +1833,18 @@ integer function itabrow6(i)
 ! all done
 
       topo%maxsystem = 5000
-      !if(mol%n.gt.500)then
-      call fragmentize(mol%n,mol%at,mol%xyz,topo%maxsystem,500,rab,topo%nb, &
-         & topo%ispinsyst,topo%nspinsyst,topo%nsystem)
-      !else
-      !   nsystem=1
-      !endif
+      !if(mol%n.gt.800)then
```
awvwgk, line 1836:
```suggestion
```

awvwgk (approved):

MtoLStoN:
Should now also raise less errors with the default hessian.

src/gfnff/frag_hess.f90
```diff
@@ -303,6 +306,11 @@ subroutine fragmentize(nspin, at, xyz, maxsystem, maxmagnat, jab, neigh, ispinsy
            end do ! End Loop over  i from 1 to count(visited)
            ass = maxloc(nspinsyst, 1)
 
+            if (nsystem .ge. nspin) then
+               call env%warning("Created more Fragments than atoms. This does not look good. Turning off fragmentization.")
```
awvwgk, line 310:
```suggestion
               call env%warning("Created more fragments than atoms. This does not look good. Turning off fragmentation.")
```

PR #656, opened 2022-06-30
Implementation of docking algorithm

cplett:
- Implementation of the intermolecular force field xtb-IFF
- General interaction site screening based on xtb-IFF

### DIFF
--- a/src/oniom.f90
+++ b/src/oniom.f90
@@ -257,6 +257,10 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, energy, gradien
          call move_alloc(gff, self%model_low)
 
       end select
+      
+      if (allocated(self%real_low%solvation)) then
+         self%model_low%solvation = self%real_low%solvation
+      end if
       call env%check(exitRun)
       if (exitRun) then
          call env%error("Could not setup low level method")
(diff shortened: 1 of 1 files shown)


meson/meson.build
```diff
@@ -39,6 +39,7 @@ elif fc.get_id() == 'intel'
       '-axAVX2',
       '-mtune=core-avx2',
       '-fma',
+      '-check all',
```
awvwgk, line 42: Please don't commit local changes to the build systems

src/set_module.f90
```diff
@@ -2385,6 +2387,18 @@ subroutine set_wall(env,key,val)
...
+subroutine set_natom(env,arg)
+   use xtb_docking_param
+   implicit none
+   type(TEnvironment), intent(inout) :: env
+   character(len=*),intent(in) :: arg
+   integer :: ios, natom
+   read(arg,'(i)',iostat=ios) natom
+   if (ios /= 0) call terminate(1)
+   natom_molA=natom
+   write(*,*) 'setting', natom_molA
```
awvwgk, line 2399: Leftover debug printout

awvwgk (changes requested):
Please add some documentation in comments before each added module and please also write at least a few rudimentary unit tests.

Issue #646, opened 2022-06-14
xTB fails to install on Windows MingW64


TyBalduf:
**Describe the bug**
I was able to build Windows in a MingW64 environment (mostly, see https://github.com/grimme-lab/xtb/issues/645). The executable in the build directory seems to run just fine. However, when I try to install the program, I get the following:
```
$ ninja -C alt_build install
ninja: Entering directory `alt_build'
[0/1] Installing files.
Installing xtb-6.dll to C:/msys64/usr/bin

ERROR: File 'xtb.lib' could not be found
FAILED: meson-install
"C:/msys64/mingw64/bin/python3.exe" "C:/msys64/home/balduf/meson/meson.py" "install" "--no-rebuild"
ninja: build stopped: subcommand failed.
```
I can't find `xtb.lib` anywhere in the build directory. I'm don't know much about the meson build process, but the issue seems to have something to do with an \IMPLIB for `xtb.lib`  not working.

**To Reproduce**
I tried to follow the [meson build instructions](https://github.com/grimme-lab/xtb/blob/main/meson/README.adoc) as closely as possible to get the directory structure to turn out right.
`meson setup alt_build  --prefix=/usr   --bindir=bin   --libdir=lib   --includedir=include   --datadir=share   --mandir=man --buildtype release --optimization 2`.

This manages to copy `xtb-6.dll` to my /usr/bin/ directory, but then fails.

TyBalduf:
I'm still looking into this, but I think the MSVC cl compiler requires a specific declaration for which functions/variables you want to export to an [import lib](https://docs.microsoft.com/en-us/cpp/build/reference/implib-name-import-library?view=msvc-170) and since these aren't (or don't seem to be) there, it doesn't make xtb.lib.

xTB can still run without this, but the various C API functions are inaccessible. I think it's possible to add a .def file to declare these functions to be exported, but I haven't yet figured out how to make meson read this file.

TyBalduf:
Adding a def seemed to do the trick. I can get the install working and am able to interface with another program. I don't know if this is an actual "bug" or just a weird configuration on my end, but linking with an external program using `-lxtb`, it expects to find `libxtb.lib` rather than `xtb.lib` which is what meson produces once a def file is included. Simply renaming the file in the install location made things work for me.

awvwgk:
It might help to add `dllexport` attributes in the C API for this cases.

[...]

PR #663, opened 2022-07-08
Export API Symbols for MSVC

TyBalduf:
* By default, the Visual Studio C compiler doesn't export any symbols,
so xtb.lib wasn't being created. Adding a def file of functions exported
and pointing to it in meson.build allows it create xtb.lib.

This should close https://github.com/grimme-lab/xtb/issues/646

### DIFF
--- a/src/api/calculator.f90
+++ b/src/api/calculator.f90
@@ -34,7 +34,6 @@ module xtb_api_calculator
    use xtb_type_molecule
    use xtb_type_calculator
    use xtb_xtb_calculator
-   use xtb_main_setup, only : addSolvationModel
    implicit none
    private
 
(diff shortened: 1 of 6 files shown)


src/api/calculator.f90
```diff
@@ -16,6 +16,8 @@
 ! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
 
 !> API for dealing with the single point calculator
+
+
```
awvwgk, line 20: Please don't commit unnecessary changes in whitespace

meson.build
```diff
@@ -108,7 +108,7 @@ xtb_lib = library(
   dependencies: lib_deps,
   include_directories: incdir,
   install: install,
-  vs_module_defs : 'xtb-6.def',
+  #  vs_module_defs : 'xtb-6.def',
```
awvwgk, line 111:
```suggestion
```

PR #705, opened 2022-10-12
Allow writing of COSMO files and usage of an ideal conductor for ddCOSMO

MtoLStoN:
Will also close #449.


src/solv/cosmo.f90
```diff
@@ -172,7 +175,7 @@ subroutine initCosmo(self, env, num, dielectricConst, nAng, radScale, vdwRad, &
    !> Offset for surface integration cutoff
    real(wp), intent(in) :: rOffset
 
-   integer :: igrid, stat, iat, izp
+   integer :: igrid, stat, iat, izp, irad
```
cplett, line 178: Is irad really used?

MtoLStoN: Seems like a coding artifact. This can be deleted.

Issue #683, opened 2022-08-25
Deviation of energies and gradients in test suite


pultar:
**Describe the bug**

Values hard-coded in the test suite do not match what the `xtb binary` produces for me with external embedding. 

https://github.com/grimme-lab/xtb/blob/1922dd31cc734edb136d3ef5ea42c7d542170e8d/test/api/c_api_example.c#L276

Deviations are relatively small, but too high to think of rounding errors. The test during build passes, however, and also my minimal example works as before. Upon more investigation, I found that gradients and energies are also inconsistent between C API and xtb binary.

**Importantly, I don't get any deviation between C API and xtb binary, when I ran a calculation without embedding (12 significant digits).** 

This might very well be an input error from my side, however, I would like to ask you to have a look. I will also open a PR with some more tests of the C API that will cover energies and gradients calculated under charge embedding conditions. 

**To Reproduce**

I compiled all files necessary to reproduce this behavior in a repo (https://github.com/pultar/xtb-example). I also attached them for your convenience:

[xtb-example-master.zip](https://github.com/grimme-lab/xtb/files/9424719/xtb-example-master.zip)

XTB binary and library are from a fresh compilation of the master branch with gfortran and openblas.

Output of my program using C API: https://github.com/pultar/xtb-example/blob/master/example.out

Output of xtb binary: https://github.com/pultar/xtb-example/blob/master/test-bohr/xtb.out

**Expected behaviour**
I expect xtb binary to produce the same energies, gradients, and point charge gradients as the C API does (and I remember, it did).

pultar:
Did you have a chance to look at my results and have any idea what is going on here? I find it quite concerning that the energy / gradients calculated differ between C API and binary. Especially the point charge gradients are off by a factor of ~20 (!)

Attached are some results as `.csv`. Could it be how I call xtb (https://github.com/pultar/xtb-example/blob/master/test-bohr/xtb.sh)?

In any case, the unit tests should be adapted accordingly to reflect this behavior.

[xtb-comparison.csv](https://github.com/grimme-lab/xtb/files/10021874/xtb-comparison.csv)

pultar:
I added two PR's, one with values generated by the C API (https://github.com/grimme-lab/xtb/pull/718) and one with values generated by the binary (https://github.com/grimme-lab/xtb/pull/719). Please note that the second test fails.

MtoLStoN:
[...]

PR #718, opened 2022-11-16
Added additional checks for C API

pultar:
In line with https://github.com/grimme-lab/xtb/issues/683, I have deviating energies, gradients, and point charge gradients depending on whether I use the C API or `xtb` binaries. 

This PR is not a solution, however, it (a) adds more tests for the C API and (b) increases the number of digits printed for PC gradients.

Values hard-coded in the tests were obtained using the C API. 

I will add a second PR with actual values obtained from the binary for easier comparison.


MtoLStoN:
The underlying issue was fixed in #720. I don't think a higher test accuracy is necessary or helpful here. But thanks for committing!

PR #721, opened 2022-11-17
Sanitize unit test and increase # significant digits PC gradients

pultar:
Thanks for the fix for https://github.com/grimme-lab/xtb/issues/683!

I agree with you (https://github.com/grimme-lab/xtb/pull/718#issuecomment-1318878837) that a third system would be overkill for the unit tests. I adjusted my PR accordingly:

- increase the number of significant digits of PC gradients in the output file to 12 to be consistent with output for gradients and energy
- the unit test for embedding / C API uses the same system as the Fortran unit test and the documentation (water tetramer) for easier comparison
- I also check in the unit test for C API for energy, and gradients with the same threshold as elsewhere

I also noted that in the test I had included earlier and that is now replaced: charges were not balanced, potentially leading to numerically less stable calculations, which is not desirable in unit tests. This PR fixes this issue.

Signed-off-by: Felix Pultar <felix.pultar@phys.chem.ethz.ch>


MtoLStoN (commented):
I agree that having consistent test accuracy is a good practice. Maybe it would be good if we also find some way to test the ``read_pcme`` subroutine, as the error that was fixed in #720 was introduced there and not caught by any of our tests.
https://github.com/grimme-lab/xtb/blob/cb55e0e4437895012fd40587f268221cbb2f437d/src/embedding.f90#L88-L91

pultar:
Sounds like a great idea! Would you like to add some code? You should be able to edit my PR.

thfroitzheim (approved):
Looks still useful to me even without the explicit test of `read_pcem`. While a more complete testing also for the cli and input/output would be good to have, we should still test point charge gradients in the meantime.

Issue #729, opened 2022-11-25
CO + water cluster segmentation fault


Viba97:
**Describe the bug**
I'm using the xtb 6.5.0 version and I'm trying to adsorb the CO molecule on a small water cluster of 66 atoms for a total of 68 atoms. However, when I try to run this calculation it gives back a segmentation fault error. It seems that error is independent to the installation (I tried on different machines) and also to the specific memory and cores requested. Despite, removing the carbon atom works fine ( 67 atoms without the C atom). I'll make my self clear in the next lines providing all the files.

These are my conditions for parallelisation and threads administration

ulimit -s unlimited
export OMP_STACKSIZE=1G
export OMP_NUM_THREADS=2,1
export OMP_MAX_ACTIVE_LEVELS=1
export MKL_NUM_THREADS=2

---FIST TEST----

**To Reproduce**
First segmentation fault error 22 H2O + CO, 68 atoms

1. xyz name --> BE_7_xyz.log
2. xtb BE_7.xyz -v --gfnff >> full_gfnff.out
3. output name --> full_gfnff_out.log

**Expected behaviour**
normal termination of xtb

-----SECOND TEST-----

After that, I tried to remove the carbon atom from the xyz file and calculation ended correctly

**To Reproduce**
67 atoms system, with 22 H2O molecules and 1 Oxygen atom

1. xyz name --> BE_7_without_carbon_xyz.log
2.  xtb BE_7_without_carbon.xyz -v -gfnff >> gfnff_without_carbon_67.out3. 
3. output name --> gfnff_without_carbon_67_out.log4. 

And as I said everything worked fine.

I do not have any clue why this is not working, any kind of help will be really appreciated.

Thank you

**Additional context**
[full_gfnff.log](https://github.com/grimme-lab/xtb/files/10093069/full_gfnff.log)
[full_gfnff_out.log](https://github.com/grimme-lab/xtb/files/10093105/full_gfnff_out.log)
[BE_7_xyz.log](https://github.com/grimme-lab/xtb/files/10093136/BE_7_xyz.log)

[gfnff_without_carbon_67_out.log](https://github.com/grimme-lab/xtb/files/10093183/gfnff_without_carbon_67_out.log)
[BE_7_without_carbon_xyz.log](https://github.com/grimme-lab/xtb/files/10093185/BE_7_without_carbon_xyz.log)

MtoLStoN:
This looks to me like a problem with how we identify Carbonyl groups here
[...]

PR #734, opened 2022-11-29
Fix GFN-FF Bug issue #729. HB Carbonyl and Nitro case.

Thomas3R:
Fixed a bug in the GFN-FF hydrogen bonding, described in issue #729


MtoLStoN (approved):
LGTM

PR #736, opened 2022-12-09
Spin-polarization in xTB

haneug:
- Simple initial commit for spin-polarization

Signed-off-by: haneug <hagen@thch.uni-bonn.de>

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1365,6 +1365,16 @@ subroutine parseArguments(env, args, inputFile, paramFile, accuracy, lgrad, &
             call env%error("No color scheme provided for --color option", source)
          end if
 
+      case('--spinpol')
+         if (get_xtb_feature('tblite')) then
+            call env%warning("Employing tblite library for spin-polarization", source)
+            call set_exttyp('tblite')
+            tblite%spin_polarized = .true.
+         else
+            call env%error("Compiled without support for tblite library. This is required for spin-polarization", source)
+            cycle
+         end if
+
       case('--oniom')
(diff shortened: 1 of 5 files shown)


src/prog/main.F90
```diff
@@ -1385,7 +1385,18 @@ subroutine parseArguments(env, args, inputFile, paramFile, accuracy, lgrad, &
          else
             call env%error("No color scheme provided for --color option", source)
          end if
-      
+
+
+      case('--spinpol')
+         if (get_xtb_feature('tblite')) then
+            call env%warning("Employing tblite library for spin-polarization", source)
+            call set_exttyp('tblite')
```
MtoLStoN, line 1393: We should only raise a warning if ``--tblite`` is not manually specified. Maybe check after evaluating all arguments if ``--spinpol`` has been set without using ``tblite``. Then you can also write a more specified warning (i.e. suggesting the user to explicitly order ``tblite`` to be used)

MtoLStoN (changes requested):

src/prog/main.F90
Albkat: Sorry, I had to transfer your check from the `newCalculator` subroutine to the `xtbMain`. Apparently, this subroutine is used for the **iff** and **docking** unit tests, and could not be modified in such way.

MtoLStoN (approved):

src/prog/main.F90
haneug: My bad, alternatively one could have checked if the optional tblite_input is present.

PR #757, opened 2023-01-18
Include SCC header with min printing

TyBalduf:
It seems like a fairly minimal amount of additional printing that can really clarify what written in the output.


MtoLStoN:
LGTM, thanks for contributing.

PR #763, opened 2023-01-30
ONIOM

Albkat:
The ONIOM routine with fixed:
1) automatic charge generation
2) interface to ORCA/TURBOMOLE
3) geometry optimization

### DIFF
--- a/src/extern/turbomole.f90
+++ b/src/extern/turbomole.f90
@@ -214,7 +214,6 @@ subroutine external_turbomole(env,n,at,xyz,nel,nopen,extcode,extmode,grd,eel,g,d
    logical :: cache
       !! to check if molecule moves b/n optimization runs
    logical :: exist
-   logical :: runCefine
    integer :: err
 
    cache = .false.
@@ -234,13 +233,16 @@ subroutine external_turbomole(env,n,at,xyz,nel,nopen,extcode,extmode,grd,eel,g,d
             if (allocated(set%ext_turbo%input_string)) then 
                call execute_command_line("exec "//cefine//" "//set%ext_turbo%input_string)
             else
-               call execute_command_line("exec "//cefine//" --func tpss --bas def2-SVP --d4 -sym c1 --noopt")
+               call execute_command_line("exec "//cefine//" --func b97-3c")
             endif
+         else
+            call env%error("No cefine binary is found", source)
+            return
          end if
 
-      !> default control file
       else
          call writeControl(chrg,uhf)
+            !! default control file
       endif
 
    end if
@@ -365,9 +367,9 @@ subroutine external_turbomole(env,n,at,xyz,nel,nopen,extcode,extmode,grd,eel,g,d
 
 end subroutine external_turbomole
 
-!----------------------------------------
-! create default TM control
-!----------------------------------------
+!-------------------------------------------------------
+! create default TM control -> future TMprep
+!-------------------------------------------------------
 subroutine writeControl(chrg,uhf)
 
    use xtb_mctc_accuracy, only : wp
--- a/src/extern/turbomole.f90
+++ b/src/extern/turbomole.f90
@@ -215,6 +215,7 @@ subroutine external_turbomole(env,n,at,xyz,nel,nopen,extcode,extmode,grd,eel,g,d
       !! to check if molecule moves b/n optimization runs
    logical :: exist
    logical :: runCefine
+   integer :: err
 
    cache = .false.
    dip=0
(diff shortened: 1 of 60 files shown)


src/prog/main.F90
```diff
@@ -478,6 +479,8 @@ subroutine xtbMain(env, argParser)
 
    ! ------------------------------------------------------------------------
    !> Print the method header and select the parameter file
+   !print*,set%runtyp
```
cplett, line 482: Some debug statements left?

src/prog/main.F90
```diff
@@ -590,10 +593,18 @@ subroutine xtbMain(env, argParser)
       select type(xtb => calc%real_low)
       type is(TxTBCalculator)
          call chk%wfn%allocate(mol%n,xtb%basis%nshell,xtb%basis%nao)
+         call newWavefunction(env,mol,xtb,chk)
```
cplett, line 596: Is it necessary here to call the `newWavefunction` subroutine every time?

src/type/anc.f90
```diff
@@ -172,7 +172,8 @@ subroutine generate_anc_blowup(self,iunit,xyz,hess,pr,linear)
 !      if (abs(self%eigv(i)) > thr1 ) elow = min(elow,self%eigv(i))
 !   enddo
 
-   damp = max(self%hlow - elow,0.0_wp)
+   damp = max(self%hlow - elow,0.0_wp) 
+   damp = damp *2
```
cplett, line 176: Why double the damping?

src/extern/turbomole.f90
```diff
@@ -163,58 +174,126 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
+         call rdvar("PATH", syspath)
+         call rdpath(syspath, "cefine", cefine, exist)
+         if (exist) then
+            call wrtm(n,at,xyz)
+            if (allocated(set%ext_turbo%input_string)) then 
+               call execute_command_line("exec "//cefine//" "//set%ext_turbo%input_string)
+            else
+               call execute_command_line("exec "//cefine//" --func tpss --bas def2-SVP --d4 -sym c1 --noopt")
+            endif
+         end if
```
MtoLStoN, line 239: Can we maybe omit the usage of cefine by writing our own control files? Something like an internal [TMPrep](https://github.com/fabothch/TMprep)? 

At least, we should probably check if ``cefine`` exists and raise an error if it doesn't (prompting the user to get it).

src/type/molecule.f90
```diff
@@ -488,29 +488,35 @@ end subroutine deallocate_molecule
 subroutine update(self)
    use xtb_mctc_accuracy, only : wp
    use xtb_pbc_tools
-   implicit none
-   class(TMolecule),intent(inout) :: self  !< molecular structure information
 
+   implicit none
```
cplett, line 492: `implicit none` is already declared for the whole module and should thus be removed here.

src/extern/turbomole.f90
```diff
@@ -163,58 +174,126 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
+               "letting TURBOMOLE take over the control..."
+            
+            if (set%oniom_settings%silent) then
+               call execute_command_line('exec ridft >  job.last 2>> /dev/null',exitstat=err)
+               if(grd)call execute_command_line('exec rdgrad >> job.last 2>> /dev/null ',exitstat=err)
+            else
+               call execute_command_line('exec ridft  | tee  job.last 2>> /dev/null',exitstat=err)
+               if(grd)call execute_command_line('exec rdgrad |tee -a job.last 2>> /dev/null ',exitstat=err)
+            endif
+
```
MtoLStoN, line 278: Same thing here - we should check if the user has Turbomole (``ridft``) before using it. Or is this done somewhere else already?

src/extern/turbomole.f90
```diff
@@ -269,9 +365,41 @@ subroutine external_turbomole(env,n,at,xyz,nel,nopen,extcode,extmode,grd,eel,g,d
 
 end subroutine external_turbomole
 
-!ccccccccccccccccccccccccccccccccc
-! TM
-!ccccccccccccccccccccccccccccccccc
+!----------------------------------------
+! create default TM control
+!----------------------------------------
```
MtoLStoN, line 370: We could use this routine to create more customizable control files (see above) - but we can also do this at a later point.

MtoLStoN (changes requested):
Can you add some tests to ensure we do not accidentally kill your ONIOM implementation with a future bugfix?

Also we should maybe enhance the handling of user induced errors (like the user having no cefine).

src/type/molecule.f90
Albkat: You are right, but always adding this statement is still good practice. It will not affect the code in any way, just some protection for the implicit typing.

src/type/anc.f90
Albkat: Yeap, forgot to delete it, thank you

src/prog/main.F90
Albkat: What do you mean by "every time"?

MtoLStoN (approved):
LGTM

PR #771, opened 2023-02-14
Charge and UHF input bug fixes for docking

cplett:


MtoLStoN (approved):
Seems like we need a Bugfix release sooner than expected ;)

PR #774, opened 2023-02-23
Fixes a bug that occurs when calculating an hessian specific builds

### LINKED ISSUE #772: --acc flag causes numerical issues and hangs
**Describe the bug**
The xtb 6.6.0 release seems to introduce some problematic behaviour with the `--acc` flag. A numerical hessian calculation may arbitrarily hang, segfault or crash with erros that numerical thresholds can not be reached.
We found this behaviour when [updating xtb in NixOS-QChem](https://github.com/Nix-QChem/NixOS-QChem/pull/250), but I can reproduce the behaviour also on Debian.
Interestingly everything works fine when the `--acc` flag is removed. The actual value of acc does not matter for the behaviour at all.

**To Reproduce**
After building on a fresh Debian OCI-Container via
```
sudo apt install gfortran cmake build-essential git libopenblas-dev
git clone https://github.com/grimme-lab/xtb.git && cd xtb
cmake -B build -DCMAKE_BUILD_TYPE=Release
make -C build
make -C build test
sudo make -C build install
```
or with Nix via `nix run github:sheepforce/nixos-qchem/xtb#xtb -- xtb.xyz --hess --input xcontrol --chrg 0 --uhf 0 --gfn 2 --acc 1.0`

1. happens with input (include input files)
`xtb.xyz`
```
4

  H -0.35738652 -2.42732105  0.01869782
  H -0.55575507  0.31681832  1.27512527
  O -0.71404094 -1.41851105  0.01869782
  O -0.20069873 -0.69255477  1.27610279
```
`xcontrol`
```
$write
            json=true
        $end
```

3. start `xtb` with (all the options here)
``` 
xtb xtb.xyz --hess --input xcontrol --chrg 0 --uhf 0 --gfn 2 --acc 1.0
```
4. run `xtb` with your options and the `--verbose` flag
5. output showing the error
[vebose.txt](https://github.com/grimme-lab/xtb/files/10734486/vebose.txt)

The behaviour is not reproducible. The very same calculation may error with the shown messages, segfault or hang infinitely from the same input with any number of OpenMP threads.

**Expected behaviour**
Reproducible behaviour, terminating calculation and correct behaviour with respect to the `--acc` flag. 

**Additional context**
The problem seems not to occur with the downloadable binary, which I assume is built with the Ifort compiler.
The BLAS implementation does not matter (tested OpenBlas, MKL and a generic BLAS).
XTB and all its dependencies pass all cmake tests.

MtoLStoN:
One of our dependencies ([mctc-lib](https://github.com/grimme-lab/mctc-lib)) handles the logic of the allocation of PBC a little bit different than we did. This lead to errors with a specific version of the gfortran compiler when explicitly copying the molecular type, that was done as a compatibility fix for ifort in #752.

Handling the PBC logical explicitly for different sized logical arrays should fix this. This will close #772.

### DIFF
--- a/src/type/molecule.f90
+++ b/src/type/molecule.f90
@@ -226,7 +226,11 @@ subroutine initMolecule &
    end if
 
    if (present(pbc)) then
-      mol%pbc = pbc
+      if (size(pbc) == 1) then
+         mol%pbc = (/pbc, pbc, pbc/)
+      else
+         mol%pbc = pbc
+      end if
       mol%npbc = count(pbc)
       if (any(pbc)) then
          mol%boundaryCondition = boundaryCondition%pbc3d


Albkat (approved):
Thank you for the fix, solves a number of issues related to the gfortran/gcc inconsistencies

PR #783, opened 2023-03-13
Fix gfnff printout for oniom and add $ignore topo 

Albkat:
1) fix oniom `gfn2:gfnff` printout  for structure optimization by printing energies and gradients during each cycle
2) add `$ignore topo` instruction to bypass the topology check when cutting bonds
3) comment: `gfnff_eg.f90`, `timer.f90`, `oniom.f90`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,4 +71,4 @@ python/build
 *.egg-info
 
 # external directory
-ext
+ext/
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,10 @@
 *.out
 *.app
 
+# Editor specific metadata
+.vscode
+*f90.swp
+
 # directories
 /build*/
 /_*/
@@ -41,7 +45,7 @@ man/*.pdf
 man/*.xml
 man/*.html
 
-# files produces by the program
+# files produced by the program
 xtbrestart
 .EDISP
 energy
@@ -65,3 +69,6 @@ __pycache__/
 python/dist
 python/build
 *.egg-info
+
+# external directory
+ext
--- a/src/gfnff/calculator.f90
+++ b/src/gfnff/calculator.f90
@@ -43,7 +43,7 @@ module xtb_gfnff_calculator
    public :: TGFFCalculator, newGFFCalculator
 
 
-   !> Calculator interface for xTB based methods
+   !> calculator interface for xTB based methods
    type, extends(TCalculator) :: TGFFCalculator
 
       type(TGFFData) :: param
@@ -70,6 +70,7 @@ contains
 
 
 subroutine newGFFCalculator(env, mol, calc, fname, restart, version)
+   
    use xtb_gfnff_param
    use xtb_gfnff_setup, only : gfnff_setup
    use xtb_disp_dftd4, only : newD3Model
@@ -100,19 +101,22 @@ subroutine newGFFCalculator(env, mol, calc, fname, restart, version)
 
    call calc%topo%zero
    calc%update = .true.
-   ! global accuracy factor similar to acc in xtb used in SCF
+
+   ! global accuracy factor similar to acc in xtb used in SCF !
    calc%accuracy = 0.1_wp
    if (mol%n > 10000) then
       calc%accuracy = 2.0_wp
    end if
 
-   !> Obtain the parameter file
+   ! obtain the parameter file !
    call open_file(ich, fname, 'r')
    exist = ich /= -1
    if (exist) then
       call gfnff_read_param(ich, calc%param)
-      call close_file(ich)
-   else ! no parameter file, try to load internal version
+      call close_file(ich) 
+      
+   ! no parameter file, try to load internal version !
+   else
       call gfnff_load_param(calc%version, calc%param, exist)
       if (.not.exist) then
          call env%error('Parameter file '//fname//' not found!', source)
@@ -134,7 +138,7 @@ subroutine newGFFCalculator(env, mol, calc, fname, restart, version)
 end subroutine newGFFCalculator
 
 
-
+!> GFN-FF wrapper for the single point energy evaluation  
 subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
       & energy, gradient, sigma, hlgap, results)
 
@@ -182,9 +186,13 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
    logical, parameter :: ccm = .true.
    logical :: exitRun
    logical :: pr
+   
+   !> GFN-FF geometry optimization
    logical :: optpr
-      !! for the case of the geometry optimization
 
+   !-------!
+   ! setup !
+   !-------!
 
    call mol%update
 
@@ -199,7 +207,7 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
       call newBornModel(self%solvation, env, solvation, mol%at)
    end if
 
-   !> To distinguish optimization, final sp and sp  
+   ! to distinguish optimization, final sp and sp ! 
    if ((set%runtyp.eq.p_run_opt).or.(set%runtyp.eq.p_run_ohess).or. &
       &   (set%runtyp.eq.p_run_omd).or.(set%runtyp.eq.p_run_screen).or. &
       &   (set%runtyp.eq.p_run_metaopt)) then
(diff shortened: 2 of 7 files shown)


src/gfnff/gfnff_eg.f90
```diff
@@ -136,8 +136,11 @@ subroutine gfnff_eg(env,pr,n,ichrg,at,xyz,makeq,g,etot,res_gff, &
      &         eeqtmp(2,n*(n+1)/2),d3list(2,n*(n+1)/2),dcn(3,n,n),cn(n), &
      &         hb_dcn(3,n,n),hb_cn(n))
 
-      if (pr) call timer%new(10 + count([allocated(solvation)]),.false.)
-
+      if (pr) then   
+         call timer%new(10 + count([allocated(solvation)]),.false.)
+      else
+         call timer%new(1, .false.)
```
awvwgk, line 142: Don't print if not verbose.

src/gfnff/gfnff_eg.f90
```diff
@@ -612,7 +615,10 @@ subroutine gfnff_eg(env,pr,n,ichrg,at,xyz,makeq,g,etot,res_gff, &
 !       which is computed here to get the atomization energy De,n,at(n)
         call goed_gfnff(env,.true.,n,at,sqrab,srab,dfloat(ichrg),eeqtmp,cn,qtmp,eesinf,solvation,param,topo)
         de=-(etot - eesinf)
+      else
+         call timer%write(6)
```
awvwgk, line 619: Is this for debugging?

src/oniom.f90
```diff
@@ -267,7 +267,7 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, energy, gradien
 
       call self%real_low%singlepoint(env, mol, chk, printlevel, restart, &
          & energy, gradient, sigma, hlgap, results)
-   
+
```
awvwgk, line 270:
```suggestion

```

awvwgk (changes requested):

src/gfnff/gfnff_eg.f90
Albkat: This one is added to print the calculation time for a single cycle during the gfnff geometry optimization.

Gfnff_opt

awvwgk: How about a *minpr* level compared to a full *pr* level of verbosity? (better variable names highly appreciated)

Albkat: What do you think about adding an optional argument to the `gfnff_eg` only for the case of optimization?
Otherwise, I do not see the point to change _pr_ into _minpr_ (or maybe _optpr_ as an alternative name).

MtoLStoN: I have to agree with @awvwgk here. I don't see this as a necessary default printout and would prefer if this would only be written if ``verbose`` is specified. 
Additionally, it would be better if we make it more clear, that this is the single cycle time. ``total time`` might be misleading.

Albkat: So i added `minpr`/`optpr `option for geometry optimization + changed `total time` to `iter.time` for clarity.
Btw, I think we  should reassign global `printlevel` var to TSet class.

Albkat: @awvwgk, I added the changes you requested. 
@MtoLStoN, the output for gfnff optimization now looks like:

image

PS: In this PR I added the commit 78fd1386db8dafc35edf001a8aba5461ddd41709 that only cleans and comments file `timer.f90`

src/gfnff/calculator.f90
```diff
@@ -196,12 +199,28 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
 
+   !> To distinguish optimization, final sp and sp  
+   if ((set%runtyp.eq.p_run_opt).or.(set%runtyp.eq.p_run_ohess).or. &
+      &   (set%runtyp.eq.p_run_omd).or.(set%runtyp.eq.p_run_screen).or. &
+      &   (set%runtyp.eq.p_run_metaopt)) then
+      if (printlevel < 2) then
+         optpr = .true.
+      else 
+         optpr = .false.
+      endif
```
awvwgk, line 210: Don't make life harder than necessary
```suggestion
      optpr = printlevel < 2
```

src/gfnff/gfnff_eg.f90
```diff
@@ -136,7 +142,13 @@ subroutine gfnff_eg(env,pr,n,ichrg,at,xyz,makeq,g,etot,res_gff, &
      &         eeqtmp(2,n*(n+1)/2),d3list(2,n*(n+1)/2),dcn(3,n,n),cn(n), &
      &         hb_dcn(3,n,n),hb_cn(n))
 
-      if (pr) call timer%new(10 + count([allocated(solvation)]),.false.)
+      if (pr) then   
+          call timer%new(10 + count([allocated(solvation)]),.false.)
+      else if (present(minpr) .and. minpr) then
```
awvwgk, line 147: You cannot check for presence of an optional argument and use the argument in the same condition as the argument might not be present and the order of evaluation is not guaranteed.

src/type/timer.f90
```diff
@@ -45,99 +47,164 @@ module xtb_type_timer
...
 
+!--------------------------------------------
+!> To initialize timer
+!--------------------------------------------
 subroutine allocate_timer(self,n,verbose)
+   
    implicit none
+   !> dummy arg list
    class(tb_timer),intent(inout) :: self
+      !! instance of timer
```
awvwgk, line 58: Please be consistent, either add all comments before or after the variable name. In `xtb` I have been using comments before the variables so far.

src/type/timer.f90
```diff
@@ -45,99 +47,164 @@ module xtb_type_timer
 
 contains
 
+!--------------------------------------------
+!> To initialize timer
+!--------------------------------------------
```
awvwgk, line 52: These are visually distracting, lets keep the comments free from those and only include the actual content
```suggestion
!> To initialize timer
```

src/type/timer.f90
```diff
@@ -45,99 +47,164 @@ module xtb_type_timer
...
 subroutine allocate_timer(self,n,verbose)
+   
    implicit none
+   !> dummy arg list
```
awvwgk, line 56: This information seems redundant.
```suggestion
```

src/type/timer.f90
```diff
@@ -45,99 +47,164 @@ module xtb_type_timer
...
    class(tb_timer),intent(inout) :: self
+      !! instance of timer
    integer, intent(in)           :: n
+      !! number of timers
    logical, intent(in), optional :: verbose
+      !! if verbose
+
+   !> local vars
```
awvwgk, line 64: Same here, I would avoid these kind of description comments
```suggestion
```

src/type/timer.f90
```diff
@@ -45,99 +47,164 @@ module xtb_type_timer
...
    real(wp) :: time_cpu
    real(wp) :: time_wall
+
    call self%deallocate
    if (n < 1) return
-   call timing(time_cpu,time_wall)
+      !! capture negative values
```
awvwgk, line 70: Both `!!` and `!>` create a docstring for tools like FORD or doxygen, these tools usually get confused if variables and statements are documented this way which are not part of the interface of the procedure.

Albkat: All right, I checked the FORD documentation, and see your point.  
I try to be consistent within the source using only predocmark (`!>`) . 

Did you try to generate xtb-docstring using FORD?

awvwgk: Yeah, I tried a while ago, but never deployed because we don't had a good coverage of docstrings so far. Still good to work toward the goal of deploying the whole docs in the future like for [tblite](https://tblite.github.io/tblite/).

Albkat: I rewrote all my comments in a consistent style, so I think this PR is good to go.

It will be helpful to establish some style guide in the best case directly in GitHub, but I am not aware if it is possible. Do you know any means of doing that? 
Another possibility would be just add [here](https://xtb-docs.readthedocs.io/en/latest/contents.html) a **developer guide**.

.gitignore
```diff
@@ -65,3 +69,6 @@ __pycache__/
 python/dist
 python/build
 *.egg-info
+
+# external directory
+ext
```
awvwgk, line 74:
```
ext/
```

awvwgk (approved):
Looks good to me now

PR #785, opened 2023-03-15
Restore support for custom lapack backend

susilehtola:
When the meson options were renamed, the support for a custom lapack backend in meson.build was also removed. This patch restores that support.


awvwgk (approved):
If nobody wants to review here, I'll have to.

PR #787, opened 2023-03-20
Add torsion potential to GFN-FF for rotation around triple bonded carbon

Thomas3R:
This potential was requested to get the correct torsion potentials for diphenylacetylene.

### DIFF
--- a/src/gfnff/gfnff_ini.f90
+++ b/src/gfnff/gfnff_ini.f90
@@ -1867,9 +1867,6 @@ subroutine gfnff_ini(env,pr,makeneighbor,mol,gen,param,topo,accuracy)
         nn = nn/2
         allocate(topo%sTorsl(6, nn), source=0)
         call specialTorsList(nn, mol, topo, topo%sTorsl)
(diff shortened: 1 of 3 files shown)


src/gfnff/gfnff_ini.f90
```diff
@@ -1849,6 +1849,30 @@ integer function itabrow6(i)
...
+        endif
+      enddo
+      if (nn.ne.0) then
+        ! fix double counting
+        nn = nn/2
+        allocate(topo%sTorsl(6, nn), source=0)
+        call specialTorsList(nn, mol, topo, topo%sTorsl)
+      else
+        ! allocate with size() = 0
+        allocate(topo%sTorsl(6, nn), source=0)
```
Albkat, line 1872: Do you really need this?
you can just remove this and then check in **gfnf_eg**, if the `topo%sTorsl` allocated

Thomas3R: I changed it below, thanks.

PR #803, opened 2023-04-12
Fix metadynamics switchon

### LINKED ISSUE #802: Metadynamics does not work with small time step
The biasing potential in metadynamics is controlled by the MD step number. This means that the biasing potential gets switched on instantaneously when a small time step is used, resulting in the system exploding. For instance, when running systems with hydrogens with `shake=0` and the usual 0.5 fs time step, one experiences an explosion already in the initial dynamics run since the starting configuration is included in the metadynamics potential.

The correct way to switch on the biasing potential is to do it as a function of actual time.

susilehtola:
Closes #802


awvwgk:
This change will have direct impact almost on all features in CREST, needs careful downstream verification before merge.

PR #807, opened 2023-04-18
Ensure that metadynamics uses slightly different initial geometry

susilehtola:
See #802

### DIFF
--- a/src/dynamic.f90
+++ b/src/dynamic.f90
@@ -345,7 +345,7 @@ subroutine md(env,mol,chk,calc, &
       if (metasetlocal%nstruc.eq.0) then
          do i = 1, mol%n
             ! Generate randomly displaced geometry
-            do while(.true.)
+            do
                rnorm = 0.0_wp
                do j = 1, 3
                   call random_number(rcoord(j))
--- a/src/dynamic.f90
+++ b/src/dynamic.f90
@@ -174,10 +174,11 @@ subroutine md(env,mol,chk,calc, &
 
    real(wp) :: step,eel1,tstep,taut,xlam2,accu,driftthr
    real(wp) :: Ekin,tmass,f,mintime
-   real(wp) :: Tinit,Tav,T,epav,ekav,dav,cma(3),bave,bavt,dum2
+   real(wp) :: Tinit,Tav,T,epav,ekav,dav,cma(3),bave,bavt
    real(wp) :: dum,edum,eerror,xx(10),molmass,slope,maxtime
    real(wp) :: tstep0,tmax,nfreedom,t0,w0,t1,w1,ep_prec,rege(4)
    real(wp) :: tors(mol%n),be(3),b0(3),tor,dip(3)
+   real(wp) :: rcoord(3), rnorm
    logical  :: ex,thermostat,restart,confdump,equi,gmd,ldum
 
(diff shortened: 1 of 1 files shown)


src/dynamic.f90
```diff
@@ -342,10 +344,23 @@ subroutine md(env,mol,chk,calc, &
       endif
       if (metasetlocal%nstruc.eq.0) then
          do i = 1, mol%n
-            do j = 1, 3
-               call random_number(dum2)
-               metasetlocal%xyz(j,i,1) = mol%xyz(j,i) + 1.0e-6_wp*dum2
+            ! Generate randomly displaced geometry
+            do while(.true.)
```
awvwgk, line 348: Just ``do`` is sufficient here as it implies an always true logical expression


```suggestion
            do
```

src/dynamic.f90
```diff
@@ -342,10 +344,23 @@ subroutine md(env,mol,chk,calc, &
...
+               rnorm = 0.0_wp
+               do j = 1, 3
+                  call random_number(rcoord(j))
+                  rnorm = rnorm + rcoord(j)**2
+               enddo
+               ! Ensure that displacement is large enough so that it
+               ! can be normalized to the unit sphere
+               if(rnorm .ge. 1e-8) then
```
awvwgk, line 356:
```suggestion
               if(rnorm >= 1e-8) then
```

src/dynamic.f90
```diff
@@ -342,10 +344,23 @@ subroutine md(env,mol,chk,calc, &
...
       if (metasetlocal%nstruc.eq.0) then
          do i = 1, mol%n
-            do j = 1, 3
-               call random_number(dum2)
-               metasetlocal%xyz(j,i,1) = mol%xyz(j,i) + 1.0e-6_wp*dum2
+            ! Generate randomly displaced geometry
+            do while(.true.)
+               rnorm = 0.0_wp
+               do j = 1, 3
+                  call random_number(rcoord(j))
```
awvwgk, line 351: I think this short loop is unnecessary if you pass the whole `rcoord` array to `random_number`

susilehtola: True, there's also an intrinsinc function for computing the norm. My FORTRAN is a bit rusty

awvwgk (approved):
Thanks for sharing, looks good to me now.

src/dynamic.f90
```diff
@@ -342,10 +344,22 @@ subroutine md(env,mol,chk,calc, &
...
+               ! Ensure that displacement is large enough so that it
+               ! can be normalized to the unit sphere
+               if(norm2(rcoord) >= 1e-8) then
+                  exit
+               endif
             enddo
+            ! Normalize displacement to the unit sphere
+            rcoord = rcoord/norm2(rcoord)
+            ! Assign displaced geometry
+            metasetlocal%xyz(j,i,1) = mol%xyz(j,i) + atom_displacement*rcoord(j)
```
awvwgk, line 362: `j` is unbound here
```suggestion
            metasetlocal%xyz(:,i,1) = mol%xyz(:,i) + atom_displacement*rcoord
```

susilehtola: Thanks, forgot to add a loop

MtoLStoN (approved):
Looks good to me. Thank you!

PR #828, opened 2023-06-27
writing updated coord AND xtbopt.xyz during external driver optimization (including nicer printout)

marcelmbn:
- write xtbopt.xyz constantly in the external driver mode so that not only "coord"-based external methods can be used for geometry optimization.
- nicer formatting

### DIFF
--- a/src/extern/driver.f90
+++ b/src/extern/driver.f90
@@ -119,18 +119,21 @@ contains
     !$omp critical (turbo_lock)
     inquire (file='gradient', exist=exist)
     if (exist) then
+      ! ### only TM output is supported for now ###
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
     end if
     if (.not. cache) then
-      call wrtm(mol%n, mol%at, mol%xyz)
-      write(env%unit,'(/,a)') &
-         "updated geometry written to: coord"
       call generateFileName(tmpname, 'xtbopt', extension, mol%ftype)
-      write(env%unit,'(a,1x,a,/)') &
+      write(env%unit,'(/,a,1x,a,/)') &
          "updated geometry written to:",tmpname
       call open_file(ich,tmpname,'w')
-      call writeMolecule(mol, ich, format=1)
+      if (exist) then
+        call writeMolecule(mol, ich, format=mol%ftype, energy=energy, &
+              & gnorm=norm2(gradient))
+      else
+        call writeMolecule(mol, ich, format=mol%ftype)
+      end if
       call close_file(ich)
 
       write (env%unit, '(72("="))')
@@ -145,6 +148,7 @@ contains
       end if
       write (env%unit, '(72("="))')
 
+      ! ### only TM output is supported for now ###
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
     end if
     !$omp end critical (turbo_lock)
--- a/src/extern/driver.f90
+++ b/src/extern/driver.f90
@@ -117,19 +117,16 @@ contains
     efix = 0.0_wp
     dipole(:) = 0.0_wp
 
-    write(*,*) printlevel
-
     !$omp critical (turbo_lock)
     inquire (file='gradient', exist=exist)
     if (exist) then
       ! ### only TM output is supported for now ###
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
-      if (cache) then
-       if (printlevel > 0) &
+      if (cache .and. printlevel > 0) then 
          write(env%unit,'(/,a,/)') &
-         "Geometry is equivalent to the one in &
-         'gradient'. Reading gradient from: 'gradient'."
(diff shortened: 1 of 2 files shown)


src/extern/driver.f90
```diff
@@ -122,6 +124,14 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
     end if
     if (.not. cache) then
       call wrtm(mol%n, mol%at, mol%xyz)
+      write(env%unit,'(/,a)') &
+         "updated geometry written to: coord"
+      call generateFileName(tmpname, 'xtbopt', extension, mol%ftype)
+      write(env%unit,'(a,1x,a,/)') &
+         "updated geometry written to:",tmpname
+      call open_file(ich,tmpname,'w')
+      call writeMolecule(mol, ich, format=1)
```
awvwgk, line 133: Why is `format=1` hard-coded here?

src/extern/driver.f90
```diff
@@ -117,11 +119,22 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
     end if
     if (.not. cache) then
-      call wrtm(mol%n, mol%at, mol%xyz)
+      call generateFileName(tmpname, 'xtbopt', extension, mol%ftype)
+      write(env%unit,'(/,a,1x,a,/)') &
+         "updated geometry written to:",tmpname
+      call open_file(ich,tmpname,'w')
+      if (exist) then
```
awvwgk, line 131: This will only write the gradient if the geometry mismatches the one for the gradient calculation

awvwgk (changes requested):

MtoLStoN (approved):

src/extern/driver.f90
```diff
@@ -117,11 +120,27 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
     !$omp critical (turbo_lock)
     inquire (file='gradient', exist=exist)
     if (exist) then
+      ! ### only TM output is supported for now ###
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
+      if (cache) then
+        write(env%unit,'(/,a,/)') &
+         "Geometry is equivalent to the one in &
+         'gradient'. Reading gradient from: 'gradient'."
```
awvwgk, line 129: Please guard this with `if (printlevel > 0)` to avoid clutter.

src/extern/driver.f90
```diff
@@ -117,11 +120,27 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
+      endif
     end if
     if (.not. cache) then
-      call wrtm(mol%n, mol%at, mol%xyz)
+      call generateFileName(tmpname, 'xtbopt', extension, mol%ftype)
+      write(env%unit,'(/,a,1x,a,/)') &
+         "updated geometry written to:",tmpname
```
awvwgk, line 136: This output should be guarded by `printlevel > 0`

src/extern/driver.f90
```diff
@@ -117,11 +120,27 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
+      if (cache) then
+        write(env%unit,'(/,a,/)') &
+         "Geometry is equivalent to the one in &
+         'gradient'. Reading gradient from: 'gradient'."
+      endif
     end if
     if (.not. cache) then
-      call wrtm(mol%n, mol%at, mol%xyz)
+      call generateFileName(tmpname, 'xtbopt', extension, mol%ftype)
```
awvwgk, line 134: We use `xtbopt` for the final optimized structure, but this routine can be called from a single point or hessian calculation. Better use a more neutral filename here.

src/extern/driver.f90
```diff
@@ -104,24 +104,48 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
+    cache = .false.
     energy = 0.0_wp
     gradient(:, :) = 0.0_wp
     sigma(:, :) = 0.0_wp
     hlgap = 0.0_wp
     efix = 0.0_wp
     dipole(:) = 0.0_wp
 
+    write(*,*) printlevel
+
```
MtoLStoN, line 121: This is probably a debug line, that should be removed.

src/extern/driver.f90
```diff
@@ -104,24 +104,48 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
...
     !$omp critical (turbo_lock)
     inquire (file='gradient', exist=exist)
     if (exist) then
+      ! ### only TM output is supported for now ###
       call rdtm(env,mol%n, .true., energy, gradient, xyz_cached)
       cache = all(abs(xyz_cached - mol%xyz) < 1.e-10_wp)
+      if (cache) then
+       if (printlevel > 0) &
```
MtoLStoN, line 129: This could be a single if statement instead of a nested one.

MtoLStoN (changes requested):

awvwgk (approved):
Looks good to me now

PR #833, opened 2023-07-11
open-shell WBOs

Albkat:
1) add Wiberg-Mayer bond orders for open-shell cases
2) add corresponding regression test (for GFN2)

### DIFF
--- a/src/scf_module.F90
+++ b/src/scf_module.F90
@@ -142,6 +142,8 @@ subroutine scf(env, mol, wfn, basis, pcem, xtbData, solvation, &
    real(wp),allocatable :: dSEdcn(:, :)
    real(wp),allocatable :: shellShift(:, :)
    real(wp),allocatable :: temp(:)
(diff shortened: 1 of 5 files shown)


src/scf_module.F90
```diff
@@ -858,11 +860,29 @@ subroutine scf(env, mol, wfn, basis, pcem, xtbData, solvation, &
...
+   !--------------------!
+   ! Wiberg bond orders !
+   !--------------------!
+   
+   ! restricted case !
+   if (wfn%nopen == 0) then
+      
+      call get_wiberg(mol%n, basis%nao, mol%at, mol%xyz,wfn%P, S, wfn%wbo,basis%fila2)
+   
+   ! unrestricted case !
```
awvwgk, line 873: That would be (restricted) open-shell rather than unrestricted. For unrestricted two spin channels are necessary

awvwgk (approved):
Looks good to me

PR #834, opened 2023-07-12
add pointer for CLI charge initialization 

Albkat:
add set%clichrg pointer to distinguish if charge is entered via CLI.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -346,7 +346,7 @@ subroutine xtbMain(env, argParser)
 
       ! Special CT input file case
       if (mol%chrg /= 0.0_wp) then
-         if (set%ichrg == 0) then
+         if (set%ichrg == 0 .and. .not. set%clichrg) then
             set%ichrg = nint(mol%chrg)
          else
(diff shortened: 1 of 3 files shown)


src/prog/main.F90
```diff
@@ -346,7 +346,7 @@ subroutine xtbMain(env, argParser)
 
       ! Special CT input file case
       if (mol%chrg /= 0.0_wp) then
-         if (set%ichrg == 0) then
+         if (set%ichrg == 0 .and. .not. set%clichrg) then
             set%ichrg = nint(mol%chrg)
          else
             call env%warning("Charge in sdf/mol input was overwritten", source)
```
MtoLStoN, line 352: ```set%ichrg==0``` seems redundant now. Do I overlook something here?

```suggestion
         if (set%clichrg) then
            call env%warning("Charge in sdf/mol input was overwritten", source)
         else
            set%ichrg = nint(mol%chrg)
```

Albkat: Yes, thank you

MtoLStoN (approved):

PR #836, opened 2023-07-26
Fixes GFN-FF bug for specific pi systems

### LINKED ISSUE #835: Different single-point GFN-FF energy with and without argument --verbose
Hello,
I'd like to report a potential bug. TOTAL ENERGY for the same structure differs with and without the --verbose flag. The version of xtb is 6.6.0.

**To Reproduce**

[P85371.zip](https://github.com/grimme-lab/xtb/files/12126591/P85371.zip)

`xtb P85371.pdb --gfnff --alpb water`
results in TOTAL ENERGY = -35.183991089038 Eh

`xtb P85371.pdb --gfnff --alpb water --verbose`
results in TOTAL ENERGY = -35.353063192998 Eh

In both cases, I also get the warning but without an idea, of how to handle it and whether is it related to the error described above.
[WARNING] Runtime exception occurred
-1- gfnff_setup: Could not read topology file.

If the error is on my part, please guide me. 
Thank you,
Ondřej

MtoLStoN:
Closes #835.

### DIFF
--- a/src/gfnff/gfnff_ini.f90
+++ b/src/gfnff/gfnff_ini.f90
@@ -926,7 +926,7 @@ subroutine gfnff_ini(env,pr,makeneighbor,mol,gen,param,topo,accuracy)
          enddo
          nelpi=nelpi-1
          Api = Apisave
-         call gfnffqmsolve(.false.,Api,S,.false.,300.0d0,npi,0,nelpi,dum,occ,eps)  !diagonalize
+         call gfnffqmsolve(.false.,Api,S,.false.,4000.0d0,npi,0,nelpi,dum,occ,eps)  !diagonalize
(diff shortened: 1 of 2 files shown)


MtoLStoN:
This bug also hits a test system. I changed the test accordingly, but we should double-check this before merging.

awvwgk (changes requested):
I think it is more useful to make this check truly informative only without affecting the outcome of the calculation.

MtoLStoN:
This would be indeed the case, if the effect on the calculation was not planned. However, I was under the assumption, that the second attempt was the intended option for this particular scenario.

MtoLStoN:
After consultation with Stefan Grimme, the effect on the outcome of the calculation seems to be indeed the intended option. However, I changed the fermi temperature to be consistent with the default case.

PR #838, opened 2023-07-31
CPCM-X

MtoLStoN:
This PR will implement the Extended Conduction-like Polarizable Continuum Solvation Model (CPCM-X) into the xTB suite. 

This should be merged only after the next Bugfix release. 

TODO: Add citation after publication.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,7 +111,7 @@ target_link_libraries(
   "$<$<BOOL:${WITH_OpenMP}>:OpenMP::OpenMP_Fortran>"
   "mctc-lib::mctc-lib"
   "$<$<BOOL:${WITH_CPCMX}>:cpcmx::cpcmx>"
-  "tblite::tblite"
+  "$<$<BOOL:${WITH_TBLITE}>:tblite::tblite>"
 )
 set_target_properties(
   "lib-${PROJECT_NAME}-static"
@@ -143,7 +143,7 @@ target_link_libraries(
   "$<$<BOOL:${WITH_OpenMP}>:OpenMP::OpenMP_Fortran>"
   "mctc-lib::mctc-lib"
   "$<$<BOOL:${WITH_CPCMX}>:cpcmx::cpcmx>"
-  "tblite::tblite"
+  "$<$<BOOL:${WITH_TBLITE}>:tblite::tblite>"
 )
 set_target_properties(
   "lib-${PROJECT_NAME}-shared"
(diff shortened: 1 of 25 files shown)


MtoLStoN:
The only thing missing should be the correct citation now after publication (in ``xhelp`` and in the ``Readme``). This should therefore be ready for Review.

awvwgk (commented):
Should we make the dependency on CPCM-X optional like the tblite one?

MtoLStoN:
We could probably do this to give the option to compile a lightweight xTB version without the additional dependencies.

MtoLStoN:
That should do the trick.

CMakeLists.txt
```diff
@@ -110,7 +110,7 @@ target_link_libraries(
   ${LAPACK_LIBRARIES}
   "$<$<BOOL:${WITH_OpenMP}>:OpenMP::OpenMP_Fortran>"
   "mctc-lib::mctc-lib"
-  "cpcmx::cpcmx"
+  "$<$<BOOL:${WITH_CPCMX}>:cpcmx::cpcmx>"
   "tblite::tblite"
```
awvwgk, line 114:
```suggestion
  "$<$<BOOL:${WITH_TBLITE}>:tblite::tblite>"
```

CMakeLists.txt
```diff
@@ -142,7 +142,7 @@ target_link_libraries(
...
```
awvwgk, line 146:
```suggestion
    "$<$<BOOL:${WITH_TBLITE}>:tblite::tblite>"
```

src/solv/cpx.F90
```diff
@@ -39,13 +69,16 @@ subroutine setup_cpcmx(self, env, solvent)
         !> TODO: SETUP solute directly from TCosmo instead of reading COSMO file
         Call read_cosmo('xtb.cosmo',self%solute,'NONE',error)
         if (allocated(error)) Call env%error(error%message, source)
+#endif
```
awvwgk, line 71: One thing to do could be adding a message that a routine was called which is not available. This should of course never happen in practice but better safe than sorry ;).

In the tblite part it should be done with something like this:
```suggestion
#else
     call NotImplementedError
#endif
```

MtoLStoN: I was actually just on it ;).

MtoLStoN:
I added two workflows that test the lightweight xTB build with cmake and meson. Seems to work pretty well.

Albkat (approved):

PR #841, opened 2023-08-01
Update Subprojects to newest stable Version.

### LINKED ISSUE #840: Dirty source release v6.6.1
Not a bug with the code, so i deleted the Issue template.

1. The [source release](https://github.com/grimme-lab/xtb/releases/download/v6.6.1/xtb-6.6.1-source.tar.xz) has pre-populated `subprojects` and a superfluous `.git` folder. The previous releases did not. 
2. The source release references a subproject version `subprojects/*.wrap` newer than in the git repo (toml-f v0.4.1, but git HEAD still has v0.2.3) . 
3. some subprojects could be updated (test-drive v0.3.0 -> v0.4.0; s-dftd3 v0.6.0 -> v1.0.0; dft-d4 v3.4.0 -> v3.5.0).

Intentional or oversight?

MtoLStoN:
Closes #840.

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -97,6 +97,82 @@ jobs:
       working-directory: ${{ env.BUILD_DIR }}
       env:
         OMP_NUM_THREADS: 2,1
+  
+  xtb-lightweight-meson-build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        fc: [gfortran-11]
+        cc: [gcc-11]
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Setup Python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.x
+ 
+    - name: Install meson
+      run: pip3 install meson==0.62.0 ninja cmake
+
+    - name: Configure build
+      run: >-
+        meson setup ${{ env.BUILD_DIR }}
+        --buildtype=debug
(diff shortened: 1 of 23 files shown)


bjoe2k4:
I've tested this on our hpc cluster (ubuntu 20.04/intel 2023.1) and it works well. 

Only downside are flaky toml-f/tftest-* tests when tests are not limited to xtb only (run tests without `--suite xtb`). These tests fails (usually just one) occur randomly when tests are run multiple times, more likely when test concurrency is increased. If i recall correctly i had this problem already with v6.6.0 but I didn't take it serious.

awvwgk:
> Only downside are flaky toml-f/tftest-* tests when tests are not limited to xtb only (run tests without --suite xtb). These test fails (usually just one) occur randomly when tests are run multiple times, more likely when test concurrency is increased. If i recall correctly i had this problem already with v6.6.0 but I didn't take it serious.

Could you report the flaky tests in TOML Fortran? https://github.com/toml-f/toml-f/issues/new/choose

bjoe2k4:
Unfortunately i cannot reproduce these flaky tests with standalone toml-f, otherwise i would have reported it upstream. Not sure how to continue from here.

MtoLStoN:
While these flaky tests are definitely something we should look at, I can also partly reproduce them with the "old" toml-f version. However, as of now, I could not see any downside except these flaky tests. Interestingly, it seems, that these hiccups only happen with the Intel Fortran Compiler and not with gfortran. 

Therefore, I believe it is safe to disregard these failing tests for the time being.

awvwgk (approved):

PR #847, opened 2023-08-15
Use "--dftbplus" instead of "--dftb+" in the thermo submodule.

### LINKED ISSUE #845: xtb thermo submodule unable to read hessian file from DFTB+
**Describe the bug**

I am trying to compute thermochemical properties using the thermo submodule; however, the program is not able to read the `hessian.out` file from DFTB+.

My `xtb` command is as follows,

`xtb thermo --sthr 100 --temp 298.15 geo_end_final.gen -dftb+ hessian.out`

I am getting the following error,

```
[ERROR] Program stopped due to fatal error
-2- Could not read hessian from 'hessian.out'
-1- io_reader_readHessian: Could not find $hessian data group
```

MtoLStoN:
It seems, the + sign in the "--dftb+" argument in the thermo submodule can lead to problems depending on the used environment.
I propose changing the default commandline argument to "--dftbplus", keeping "--dftb+" as an alternative for backward compatibility.

Closes #845.

### DIFF
--- a/src/prog/thermo.f90
+++ b/src/prog/thermo.f90
@@ -204,7 +204,7 @@ subroutine parseArguments(env, args, htype, massWeighted)
          call thermoHelp(env%unit)
          call terminate(0)
 
-      case('--dftb+')
+      case('--dftb+','--dftbplus')
          htype = hessType%dftbplus
 
       case('--turbomole')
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Looks good to me.

PR #851, opened 2023-08-21
update ONIOM and TAtomList

Albkat:
1) add warning if the distances between added link atom (H)  and any other atom is too short (smaller than H2) -> `check_dist`

2) update man page

3) allow for **0**  in TAtomList (e.g. `1,2,0,12`)

### DIFF
--- a/man/xcontrol.7.adoc
+++ b/man/xcontrol.7.adoc
@@ -336,10 +336,14 @@ $oniom
 *derived k*='bool'::
    to calculate prefactor *k* and create jacobian dynamically (see more )
 
+*ignore topo*='bool'::
+   to bypass topology check when breaking bonds
+
+*outer*='bool'::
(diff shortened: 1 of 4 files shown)


awvwgk:
Why do we need zeros in the atom list, this sounds like some out of bounds issue to me.

awvwgk:
I think we should rather add checks to avoid adding zeros in the first place to the list.

Albkat:
> I think we should rather add checks to avoid adding zeros in the first place to the list.

I assume that performing a direct check is not possible since the **TAtomList** parser is explicitly composed of pure procedures. I've altered the error handling approach: now, whenever it encounters a non-integer value or zero, it throws an error. If you have any more insightful suggestions for enhancing this, I'd be glad to hear them.

awvwgk (approved):
Looks good to me then

PR #853, opened 2023-08-23
add ONIOM references and update derived mode

Albkat:
update `oniom.f90` w.r.t. https://doi.org/10.1039/D3CP02178E publication


MtoLStoN (approved):

awvwgk (commented):
I think the reference should be part of the Readme and output as well.

MtoLStoN (approved):

PR #864, opened 2023-09-13
Add "--tmcosmo" mode for writing .cosmo files with TM convention.

### LINKED ISSUE #863: COSMO Screening Charge and Volume
I have encountered two issues when using XTB COSMO outputs with other COSMO related programs, surface charges and volume. 

After performing some XTB COSMO calculations and comparing them with DFT calculations, I have noticed that the COSMO surface charges appear to be for the molecule charge and not the screening charge. For example in the case of water, the surface charge near oxygen is negative but the screening charge of the solvent should be positive (see attached). It would be useful if there was an option to switch the COSMO output from molecule to screening charge, as is convention. This would allow other programs which use COSMO files to easily use XTB COSMO output.

The other issue is that while XTB outputs an area during the COSMO calculation, other COSMO programs output a volume which can be used in further calculations. It would be useful if volume was included in the XTB COSMO output.

Thanks,
Angus

XTB_COSMO_Water_Surface SigmaProfile

MtoLStoN:
Related to #863. Uses the TM convention for the writing of ``.cosmo`` files in xTB when the "--tmcosmo" keyword is used.

Edit: Given the scope of this PR, this will now completely close #863 if merged.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1576,11 +1576,11 @@ subroutine parseArguments(env, args, inputFile, paramFile, accuracy, lgrad, &
             call env%error("No solvent name provided for ALPB", source)
          end if
 
-      case('--cosmo')
+      case('--cosmo','--tmcosmo')
          call args%nextArg(sec)
          if (allocated(sec)) then
             call set_gbsa(env, 'solvent', sec)
-            call set_gbsa(env, 'cosmo', 'true')
+            call set_gbsa(env, flag(3:), 'true')
             call args%nextArg(sec)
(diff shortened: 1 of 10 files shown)


Albkat (commented):
Maybe it makes sense to pass the `set%solvInput%tmcosmo` value to `writeCosmoFile` routine. This way you can get rid of the optional attribute, as well as redundant `tm` variable.

Albkat (commented):
Nicely done!
It might be worthwhile to consider adding a regression test for this.

Albkat (approved):

PR #868, opened 2023-09-18
Introducing a scaling factor for directed docking

cplett:
The additional attractive/repulsive potential for the directed docking feature of the `aISS` algorithm can now be scaled.
Therefore, the `scaling factor= <value>` keyword of the `$directed` block in the additional input file is used.


Albkat (commented):
what is the difference between `rdblock_docking` and `rdblock_docking2`?

MtoLStoN (approved):

PR #869, opened 2023-09-20
Implement dipole moment in GFN-FF

Thomas3R:
The GFN-FF force field is now able to calculate dipole moments from EEQ charges and atom positions.


MtoLStoN (approved):

Albkat (commented):
Do you plan to extend `gfnff_property` routine in the future? Otherwise, this seems redundant to me.

PR #906, opened 2023-11-20
specify ifort location

Albkat:
Apparently, in the new oneAPI version, the  **setvars.sh** script does not configure `ifort`.

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -263,6 +263,7 @@ jobs:
       run: |
         sudo apt-get install ${APT_PACKAGES}
         source /opt/intel/oneapi/setvars.sh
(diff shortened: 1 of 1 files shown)


awvwgk:
Can we check whether we are happy with the fix compiler instead of fort, since the latter is now slowly getting deprecated by Intel?

Albkat:
As a simple trick with the name switching does not work,  I have opened an issue #908. 
I would suggest pushing this PR anyway, because the `ifx` is also not configured by the **setvars.sh**, and one has to explicitly source the corresponding **vars.sh**.

MtoLStoN:
We should definitely look into transitioning to the ``ifx`` compiler before it is deprecated in 2024, but I think, we can use this as an intermediate solution.

MtoLStoN (approved):

PR #907, opened 2023-11-21
resolve debug compilation errors

Albkat:
CMake build in debug mode results in several runtime errors (with both `gfortran` v11.4.0 and `ifort` v2021.5.0):
1. Race condition in `dftd4.f90`
2. Out of bound exception in `oniom.f90` 
3. Dimension mismatch in `molecule.f90`


Albkat:
In addition, I tested the Meson build with in the debug mode. Unfortunately, the resulting binaries do not arise any exceptions during the runtime. It might be advisable to consider rewriting the `meson.build` script in the future.

Albkat:
Big thanks to @MtoLStoN and @marvinfriede  for their assistance during the debugging process.

MtoLStoN (approved):

Issue #819, opened 2023-05-26
Using external program with xTB 6.6.0


amavr78:
In xtb 6.6.0 , the program crashes with a segmentation fault , when I try to call ORCA as an external program with xTB .
This is during a calculation with censo, but the error is due to the xtb .
This is not happening with xtb 6.5.1

I am trying to execute with the command:  xtb coord --orca -I opt.inp 

The contents of the opt.inp are:
$opt
maxcycle=8
microcycle=8
average conv=true
hlow=0.01
s6=30.00
engine=lbfgs
$external
   orca input file= inp
   orca bin= /data1/bin/ORCA/orca_5_0_4_linux_x86-64_openmpi411/orca
$end

When  I execute , I get the following segmentation fault:
           -------------------------------------------------
          |                Calculation Setup                |
           -------------------------------------------------

          program call               : xtb coord --orca -I opt.inp
          coordinate file            : coord
          omp threads                :                    40

   ID    Z sym.   atoms
    1    6 c      1-4, 6-9, 11-13, 20, 25, 28, 31, 34, 35, 38
    2    7 n      5, 10
    3    1 h      14-19, 21-24, 26, 27, 29, 30, 32, 33, 36, 37, 39, 40, 42,
                  43, 45, 46
    4   16 s      41, 44
    5    8 o      47-52
forrtl: severe (174): SIGSEGV, segmentation fault occurred
Image              PC                Routine            Line        Source             
xtb                000000000046D41A  Unknown               Unknown  Unknown
xtb                00000000037F12A0  Unknown               Unknown  Unknown
xtb                0000000000D05C22  xtb_extern_orca_m         198  orca.f90
xtb                00000000006B39A3  xtb_main_setup_mp         160  setup.f90
xtb                000000000042C7EE  xtb_prog_main_mp_         526  main.F90
xtb                0000000000450642  MAIN__                     57  primary.f90
xtb                00000000004067F2  Unknown               Unknown  Unknown
xtb                00000000037F2740  Unknown               Unknown  Unknown
xtb                00000000004066D7  Unknown               Unknown  Unknown

I am using the precompiled binaries.
I thought that the error could be related to the optimization , but it is not. 
It is crashing, even If I ask to do an energy calculation.

The xtb 6.5.1 is running fine without any errors.

MtoLStoN:
[...]

PR #916, opened 2023-11-25
refactor ORCA calculator

Albkat:
update all ORCA-related code:

1. enhance the readability of `orca.f90` and `driver.f90 `
2. add `checkExe` to validate the external binary 
3. add `chkInput` and `firstArgComesFirst` to check for ORCA input for certain keywords and their order
4. rewrite `writeOrcaInp` 
5. add an extra level of error handling (`exitstat` does not catch **OpenMPI**-related ORCA crashes )

### DIFF
--- a/src/extern/driver.f90
+++ b/src/extern/driver.f90
@@ -15,111 +15,130 @@
 ! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
 
 module xtb_extern_driver
-  use xtb_mctc_accuracy, only: wp
-  use xtb_mctc_io, only: stdout
-  use xtb_mctc_filetypes, only: fileType, generateFileName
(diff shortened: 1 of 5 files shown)


Albkat:
In addition, I tested xtb in the wrapper mode for the ORCA to avoid issues like this #819.

marcelmbn (approved):
Looks good to me!
Tested the following functionalities on a sample SH3+ structure:
- `--driver "numgradpy <args>"`
- `--tm`
- `--orca`

... and everything worked smoothly as expected. 

However, for future PRs, I'd suggest decoupling formatting and "actual" code changes to keep it readable within the `diff` concept. (So that one can easily identify the relevant changes in the code so that not everything's green and red.)

PR #917, opened 2023-11-27
ONIOM results

Albkat:
add ONIOM specific property printout

Thanks @benbaed for the hint.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -958,6 +958,9 @@ subroutine xtbMain(env, argParser)
    type is(TxTBCalculator)
       call write_energy(env%unit,res,fres, &
         & (set%runtyp.eq.p_run_hess).or.(set%runtyp.eq.p_run_ohess).or.(set%runtyp.eq.p_run_bhess))
+   type is(TOniomCalculator)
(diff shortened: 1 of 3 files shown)


src/prog/main.F90
```diff
@@ -958,6 +958,9 @@ subroutine xtbMain(env, argParser)
    type is(TxTBCalculator)
       call write_energy(env%unit,res,fres, &
         & (set%runtyp.eq.p_run_hess).or.(set%runtyp.eq.p_run_ohess).or.(set%runtyp.eq.p_run_bhess))
+   type is(TOniomCalculator)
+      call write_energy_oniom(env%unit,res,fres, &
+         & (set%runtyp.eq.p_run_hess).or.(set%runtyp.eq.p_run_ohess.or.(set%runtyp.eq.p_run_bhess)))
```
MtoLStoN, line 963: This has no effect on functionality of course, but is there a specific reason why you added the brackets like that?

MtoLStoN (approved):

src/prog/main.F90
Albkat: nah, just a typo

PR #921, opened 2023-12-04
add CMake option INSTALL_MODULES

njzjz:
This option has been in the meson but not in the CMake.


awvwgk (approved):

PR #928, opened 2023-12-12
constrain all atoms

Albkat:
add option to constrain all atoms:
```
$constrain
    atoms: all``

### CODE REVIEW

File: src/constrain_param.f90 [hunk 1]
```diff
@@ -452,12 +452,19 @@ subroutine set_constr(env,key,val,nat,at,idMap,xyz)
       potset%pos%n = size(list)
    case('atoms')
       call atl%new(val)
```
awvwgk, line 454: Should this only be done if the value is not all?

MtoLStoN (approved):

PR #937, opened 2023-12-22
temporary fix for CI

Albkat:
Address the issue of the new ifort version not compiling programs by temporarily reverting CI to an older version of ifort as a workaround, until ifx is fully integrated.


awvwgk (approved):

awvwgk:
Maybe also add a hint for the upper bound of ifort in the README

Albkat:
> Maybe also add a hint for the upper bound of ifort in the README

Thanks, the hint is added. 
Since I don't have access to any CUDA machines, I still lack information on the PGI build. However, this should be irrelevant for the current pull request.

PR #940, opened 2023-12-26
read ml params from file

dpooria:


awvwgk (changes requested):
This needs some more work, especially cleanup regarding debug printouts, duplicated files and comments for documentation. Also, the motivation for this is not clear since there is an existing parameter format which can be read from a file.

dpooria:
Sorry for the mistake. This is a work under development and the PR was meant to be on our fork not the main repo.

PR #947, opened 2024-01-09
refactor geometry optimization (part 1)

Albkat:
1.  Implement partial optimization of the `xtbMain` subroutine.
2.  Add detailed comments to the `xtb_geoopt` module to enhance code clarity
3.  Introduce the **tsopt.f90** file for the forthcoming Transition State (TS) optimizer.

### DIFF
--- a/src/geoopt_driver.f90
+++ b/src/geoopt_driver.f90
@@ -18,65 +18,80 @@
 module xtb_geoopt
 contains
 
+!> wrapper for relaxation engines
+!> more info: https://xtb-docs.readthedocs.io/en/latest/optimization.html 
 subroutine geometry_optimization &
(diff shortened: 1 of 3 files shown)


MtoLStoN:
Tbh, I do not understand, why we would need to introduce an empty module file in this commit? Could you explain your reasoning?

Albkat:
> Tbh, I do not understand, why we would need to introduce an empty module file in this commit? Could you explain your reasoning?

Yeah, my intention was to initially include this file as an empty one progressively developing a TS optimizer. However, I understand the point for doing this in a separate PR and in a more finalized state, which is why I have removed the corresponding commit.

MtoLStoN (approved):

PR #956, opened 2024-01-26
aISS docking update for next QCG version

cplett:
Introduction of an optional attractive potential during the position screening of the `aISS` algorithm. This will be used to make the solute more attractive for QCG runs. It has the shape of an error function including three parameters that can be used to determine the strength, the width, and the turning point. They are provided via an input file and stored in the `directedset` object.

### DIFF
--- a/src/docking/set_module.f90
+++ b/src/docking/set_module.f90
@@ -245,11 +245,15 @@ contains
       integer, allocatable :: list(:)
       real(wp) :: ddum
       logical  :: ldum
+      real(wp), parameter ::au = 627.509541d0
    
       integer  :: narg
       character(len=p_str_length),dimension(p_arg_length) :: argv
 
       logical, save :: set1 = .true.
+      logical, save :: set2 = .true.
+      logical, save :: set3 = .true.
+      logical, save :: set4 = .true.
    
       call atl%resize(nat)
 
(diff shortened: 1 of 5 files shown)


src/docking/set_module.f90
```diff
@@ -245,11 +245,15 @@ subroutine set_directed(env,key,val,nat,at,idMap,xyz)
       integer, allocatable :: list(:)
       real(wp) :: ddum
       logical  :: ldum
+      real(wp), parameter ::au = 627.509541d0
```
Albkat, line 248:
```suggestion
      use xtb_mctc_convert, only : autokcal
```

src/docking/set_module.f90
```diff
@@ -303,6 +307,16 @@ subroutine set_directed(env,key,val,nat,at,idMap,xyz)
...
+      case('expo')
+         if (getValue(env,val,ddum).and.set2) directedset%expo(1) = ddum
+         set2 = .false.
+      case('prefac')
+         !Prefactor is given in kcal, but xtb energies are in hartree
+         if (getValue(env,val,ddum).and.set3) directedset%val(1) = ddum / au
+         set3 = .false.
+      case('midpoint')
+         if (getValue(env,val,ddum).and.set4) directedset%expo(2) = ddum
+         set4 = .false.
```
Albkat, line 319: Since these instructions aren't documented in [xcontrol.7.adoc](https://github.com/grimme-lab/xtb/blob/main/man/xcontrol.7.adoc), it might be beneficial to add them to the [xtb-docs](https://xtb-docs.readthedocs.io/en/latest/xtb_docking.html)

Albkat (commented):
Looks good

Albkat (dismissed):

Albkat (approved):

PR #960, opened 2024-02-02
Add charge and uhf information into setup

Albkat:


src/scf_module.F90
```diff
@@ -479,6 +479,8 @@ subroutine scf(env, mol, wfn, basis, pcem, xtbData, solvation, &
       write(env%unit,scifmt) "-> SCF convergence ",scfconv, "Eh  "
       write(env%unit,scifmt) "-> wf. convergence ",qconv,   "e   "
       write(env%unit,dblfmt) "Broyden damping    ",set%broydamp,"    "
+      write(env%unit,intfmt) "net charge         ",set%ichrg
```
awvwgk, line 482: Is this or mol%chrg actually used?

Albkat: yeah, in principle `mol%chrg` is set based on the  `set%ichrg` value, but it might be more logical to retrieve this information from `mol` directly.

Albkat: Thank you for the suggestion; I've updated the PR accordingly.

awvwgk (approved):

src/scf_module.F90
lizhiq16: how about the uhf information?

PR #967, opened 2024-02-19
raman activities with adjustable wavelength and T

benbaed:
add `ptbsetup` variables `raman_temp` and `raman_lambda`

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -231,14 +231,20 @@ contains
 
       !> If hessian (or ohess or bhess) is requested in combination with PTB, conduct GFN2-xTB + PTB hessian
       anyhess = (set%runtyp == p_run_hess) .or. (set%runtyp == p_run_ohess) .or. (set%runtyp == p_run_bhess)
-      if (set%mode_extrun == p_ext_ptb .and. anyhess) then
-         set%mode_extrun = p_ext_xtb
-         set%ptbsetup%ptb_in_hessian = .true.
-         call set_gfn(env, 'method', '2')
-         call set_gfn(env, 'd4', 'true')
-         tblite%method = "gfn2"
-         set%ptbsetup%hessmethod = "GFN2-xTB"
-      end if
+      if (anyhess) then
+         if(set%mode_extrun == p_ext_ptb) then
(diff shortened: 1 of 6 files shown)


src/prog/main.F90
```diff
@@ -1778,6 +1778,18 @@ subroutine parseArguments(env, args, inputFile, paramFile, lgrad, &
          case ('--alpha')
             call set_elprop('alpha')
 
+         case ('--raman')
+            call set_elprop('raman')
```
marcelmbn, line 1782: In the current state, only the electrical property is set to `alpha` (via redirect `raman`), but no Hessian calculation is conducted. I.e., no frequencies (and intensities) will be available.
One could add this easily via a call similar to: `call set_runtyp('hess')`.

marcelmbn, line 1782: Similar to the comment above, a switch to turn on PTB should be added:
```
            call set_exttyp('ptb')
            if (.not. get_xtb_feature('tblite')) then
               call ptb_feature_not_implemented(env)
               return
            endif
```

marcelmbn: The intended principle behind `--raman` is in my opinion the combination of the flags `--ptb --hess --alpha` together with Raman-specific options like temperature and frequency of incident light.

marcelmbn: (The automatic redirect to `--ptb` follows from the fact that PTB is currently the only implemented method capable of calculating polarizabilities.)

marcelmbn (changes requested):
Inclusion of "`--hess`" and "`--ptb`" calls directly into `--raman` would make sense. Other than that, this looks good to me.

src/prog/main.F90
Albkat: Good idea, but I'd still preserve `--ptb` as a requirement in case of the Raman implementation with any other tight-binding method.

marcelmbn: Also fine for me. However, it should be checked whether an error is produced if a call like `--raman --gfn 2` or similar is used. (Something like: "Raman not implemented with this method".)

marcelmbn (approved):

marcelmbn:
Looks good to me now!

PR #969, opened 2024-02-23
use accu_hess

Albkat:
set higher SCC accuracy for the numerical hessian calculations (variable `set%accu_hess` has already been defined, but not used).

### DIFF
--- a/src/hessian.F90
+++ b/src/hessian.F90
@@ -176,7 +176,7 @@ subroutine numhess( &
    call axis(mol%n,mol%at,mol%xyz,aa,bb,cc)
    if(cc.lt.1.d-10) res%linear=.true.
    step2=0.5_wp/step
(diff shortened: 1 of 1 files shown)


src/hessian.F90
```diff
@@ -176,6 +176,7 @@ subroutine numhess( &
    call axis(mol%n,mol%at,mol%xyz,aa,bb,cc)
    if(cc.lt.1.d-10) res%linear=.true.
    step2=0.5_wp/step
+   calc%accuracy=set%accu_hess
```
awvwgk, line 179: Should this be reset after the hessian calculation?

Albkat: Nica catch, thanks!  (ceb0d22)

awvwgk (approved):

PR #972, opened 2024-02-28
add Stale workflow

Albkat:
Automatically close PRs and issues labeled as **stale** if there's no activity within 7 days.


MtoLStoN:
If I understanding correctly, this will not automtically add stale labels but only closed PRs that are already labeled as stale. Do you want to add the stale labels manually?

MtoLStoN (approved):
Looks good to me then

PR #985, opened 2024-03-07
add json output for tblite

Albkat:


src/prog/main.F90
```diff
@@ -1020,32 +1018,39 @@ subroutine xtbMain(env, argParser)
...
-         select type (calc)
+         select type (calc)         
          type is (TxTBCalculator)
             call open_file(ich, 'xtbout.json', 'w')
             call main_xtb_json(ich, &
                                mol, chk%wfn, calc%basis, res, fres)
             call close_file(ich)
+
+         type is (TTBLiteCalculator)
+            call open_file(ich, 'tblite.json', 'w')
```
awvwgk, line 1029: Maybe it would be better to try to match the format of xtbout.json here

Albkat: We could print both:)

PR #991, opened 2024-03-13
Add dependecies in meson build

Thomas3R:


Albkat (approved):
lgtm

TyBalduf:
@Thomas3R @Albkat 
Are these required dependencies? Its not immediately clear looking at the commits between v6.7.0 and here what led to a requirement for these packages.

If these are not required, it would be useful to have an option similar to cpx/tblite to enable/disable them. I noticed a build failure locally on my Windows machine (intel-cl 2021.5.0 Fortran/(msvc 19.29.30141 C compilers). I could probably patch around, but if they really are optional, I could put together a pull request to make them optional.

PR #994, opened 2024-03-14
Reduce memory usage in egbond_hb for GFN-FF

Thomas3R:
The matrix considered_ABH(n,n,n) allocation used in the HB scaled bond energy was unnecessarily large. I introduced a mapping from the atom index to the hb index for AB atoms and H atoms separately. Since this is done in the initialization of the force field, the new variables need to be part of the topology as well.


Albkat (approved):
Thanks for sharing! LGTM.

PR #964, opened 2024-02-13
reformat files relevant for PTB PR with fprettify

marcelmbn:
Files that are relevant for the PTB PR https://github.com/grimme-lab/xtb/pull/963 were consistently reformatted using `fprettify` with the following settings:

```python
    "fortran.formatting.fprettifyArgs": [
        "--whitespace-plusminus",
        "--whitespace-logical",
        "--whitespace-multdiv",
        "--enable-replacements",
        "--whitespace-intrinsics",
        "--whitespace-comma",
        "-l 120",
        "--enable-decl",
        "--c-relations",
    ],
```

PR #1004, opened 2024-03-27
implemented -bhess option in xtb thermo submodule

gorges97:
Frequencies of the hessian file need to be scaled to remove the frequency shift caused by the bias RMSD potential. With the --bhess flag, xtb thermo now reads in an hessian_sph file from a previous xTB SPH calculation and scales the frequencies accordingly to reproduce the thermo contributions of the xTB SPH calculation.

### DIFF
--- a/src/hessian.F90
+++ b/src/hessian.F90
@@ -728,7 +728,7 @@ subroutine rescale_freq(n3,htb,hess,hbias,freq,fc_tb,fc_bias,freq_scal)
    real(wp), intent(out) :: freq_scal(n3) 
    real(wp),allocatable :: v(:)
    real(wp),allocatable :: fc_tmp(:)
-   real(wp), parameter :: alp1=1.27_wp, alp2=1.5d-4
+   real(wp), parameter :: alp1=1.27_wp, alp2=1.5e-4_wp
    integer, intent(in) :: n3
    integer :: j
    
@@ -740,9 +740,9 @@ subroutine rescale_freq(n3,htb,hess,hbias,freq,fc_tb,fc_bias,freq_scal)
       fc_tb(j) = mctc_dot(v,fc_tmp)
       call mctc_gemv(hbias,v,fc_tmp)
       fc_bias(j) = mctc_dot(v,fc_tmp)
-      if (abs(freq(j)).gt.1.0d-6) then
+      if (abs(freq(j)) .gt. 1.0e-6_wp) then
          freq_scal(j) = sqrt( (fc_tb(j)+alp2) / ( (fc_tb(j)+alp2) +  alp1*fc_bias(j) ) )
-         if (fc_tb(j).lt.0.and.fc_bias(j).ne.0) then
+         if (fc_tb(j) .lt. 0.0_wp .and. fc_bias(j) .ne. 0.0_wp) then
             freq_scal(j) = -sqrt( (abs(fc_tb(j))+alp2) / ( (abs(fc_tb(j))+alp2) + alp1*fc_bias(j) ) )
          end if
       else
--- a/src/hessian.F90
+++ b/src/hessian.F90
@@ -355,6 +355,14 @@ subroutine numhess( &
    write(env%unit,'("writing file <",a,">.")') hname
    call wrhess(n3,hss,hname)
 
+   ! non mass weigthed biased Hessian in hsb
+   if (set%runtyp .eq. p_run_bhess) then
+      hname = 'hessian_sph'
+      write (env%unit, '(a)')
+      write (env%unit, '("writing file <",a,">.")') hname
+      call wrhess(n3, hsb, hname)
+   end if
+
    ! include masses
    k=0
    do i=1,n3
--- a/src/hessian.F90
+++ b/src/hessian.F90
@@ -25,7 +25,7 @@ module xtb_hessian
 
    public :: numhess
    public :: trproj, rdhess, g98fake2, distort, write_tm_vibspectrum
-
+   public :: rescale_freq
 contains
 
 subroutine numhess( &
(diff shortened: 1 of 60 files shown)


marcelmbn:
Why do we need `xtb thermo` at all for `--bhess`? When you run `xtb struc.xyz --bhess` you already get all relevant information, don't you?

gorges97:
The xtb thermo module can save you the computation time to recompute the hessian. At the moment, xtb thermo gives wrong thermostatistical contributions if it is used for the hessian of an SPH calculation.

src/prog/thermo.f90
```diff
@@ -164,8 +174,75 @@ subroutine preigf(unit, freq, scale)
...
+      call massWeightHessian(hessian_sph, atmass) ! keep units similar to hessian.f90 routines in main program
+      call massWeightHessian(hessian, atmass) ! keep units similar to hessian.f90 routines in main program
+      htb=hessian-hessian_sph
+
+      call diagHessian(env, hessian, freq)
+
+      allocate(v(n3),fc_tmp(n3),freq_scal(n3),fc_tb(n3),fc_bias(n3))
+      ! calculate fc_tb and fc_bias (same procedure as in hessian.f90)
+      alp1=1.27_wp
+      alp2=1.5d-4
```
marcelmbn, line 213: Can we avoid using fixed float point numbers?

marcelmbn: Or is the whole function "copied" from the `bhess` workflow?

gorges97: I just "copied" the function from the bhess workflow. Otherwise the bhess subroutine has to be also rewritten to avoid this "double code".

marcelmbn: Can't we just use it?

marcelmbn: Can you highlight which lines of code from `src/hessian.f90` are actually relevant and could be "recycled" here? So that we can think about putting these parts of the code into single functions that could be reused in this project. I'd strongly speak against implementing doubled code using fixed float point parameters (which is in itself already a bit of bad practice), if we can avoid it.

TyBalduf: It looks like it is mainly lines 397-416 from the updated hessian.F90 that could be reused here fairly easily.

src/hessian.F90
```diff
@@ -721,6 +715,43 @@ subroutine distort(mol,freq,u)
...
+   implicit none
+   real(wp),intent(in) :: htb (n3,n3)
+   real(wp),intent(in) :: hess(n3,n3)
+   real(wp),intent(in) :: hbias(n3,n3)
+   real(wp),intent(in) :: freq(n3)
+   real(wp), intent(out) :: fc_tb(n3),fc_bias(n3)
+   real(wp), intent(out) :: freq_scal(n3) 
+   real(wp),allocatable :: v(:)
+   real(wp),allocatable :: fc_tmp(:)
+   real(wp), parameter :: alp1=1.27_wp, alp2=1.5d-4
```
marcelmbn, line 731:
```suggestion
   real(wp), parameter :: alp1=1.27_wp, alp2=1.5e-4_wp
```

src/hessian.F90
```diff
@@ -721,6 +715,43 @@ subroutine distort(mol,freq,u)
...
+   
+   allocate(fc_tmp(n3),v(n3))
+   ! calculate fc_tb and fc_bias
+   do j=1,n3
+      v(1:n3) = hess(1:n3,j) ! modes
+      call mctc_gemv(htb,v,fc_tmp)
+      fc_tb(j) = mctc_dot(v,fc_tmp)
+      call mctc_gemv(hbias,v,fc_tmp)
+      fc_bias(j) = mctc_dot(v,fc_tmp)
+      if (abs(freq(j)).gt.1.0d-6) then
```
marcelmbn, line 743:
```suggestion
      if (abs(freq(j)) .gt. 1.0e-6_wp) then
```

src/hessian.F90
```diff
@@ -721,6 +715,43 @@ subroutine distort(mol,freq,u)
...
+         freq_scal(j) = sqrt( (fc_tb(j)+alp2) / ( (fc_tb(j)+alp2) +  alp1*fc_bias(j) ) )
+         if (fc_tb(j).lt.0.and.fc_bias(j).ne.0) then
```
marcelmbn, line 745:
```suggestion
         if (fc_tb(j) .lt. 0.0_wp .and. fc_bias(j) .ne. 0.0_wp) then
```

src/prog/thermo.f90
```diff
@@ -164,8 +174,60 @@ subroutine preigf(unit, freq, scale)
...
+      
+      call rescale_freq(n3,htb,hessian,hessian_sph,freq,fc_tb,fc_bias,freq_scal)
+
+      ! scale frequencies and convert now to atomic units
+      do j=1,n3
+         freq(j) = freq(j) * freq_scal(j)/sqrt(amutoau) 
+      end do
+
+      call preigf(env%unit, freq, autorcm)
+	else
```
marcelmbn, line 221: Make format more consistent, think about using `fprettify`, as described here:
https://github.com/grimme-lab/xtb/pull/964#issue-2131836030

src/prog/thermo.f90
```diff
@@ -164,8 +174,60 @@ subroutine preigf(unit, freq, scale)
 end subroutine preigf
 
 
+! Remove the frequency shift caused by the bias RMSD employed in the SPH calculation:
+subroutine remove_sphshift(env,argParser,mol, hessian, freq)
+   use xtb_hessian
+    real(wp), allocatable, intent(inout) :: freq(:), hessian(:, :)
```
marcelmbn, line 180: Is the `allocatable` statement at this point actually required when allocation happened already before?

src/prog/thermo.f90
```diff
@@ -105,20 +111,24 @@ subroutine xtbThermo(env, argParser)
...
-         call massWeightHessian(hessian, mol%atmass)
-      end if
-
-      if (hFormat == hessType%dftbplus .or. hFormat == hessType%orca) then
-         call projectHessian(hessian, mol, .true., .true.)
+      
+      if (bhess) then
+         call remove_sphshift(env,argParser,mol,hessian,freq) ! frequencies have to be rescaled 
+      else 
+        if (.not.massWeighted) then
```
marcelmbn, line 118: Format comment also applies at these points of the code... 😉😄

marcelmbn (changes requested):

PR #1008, opened 2024-04-05
include preprocessor definitions 

Albkat:
include  the preprocessor flags during regression tests in the Meson build.


awvwgk (approved):
Good catch

PR #1010, opened 2024-04-12
add --ceh flag

Albkat:
calculate CEH charges via tblite interface.
Small summary:
1. `get_ceh` procedure to obtain the charges from the library.
2. write the charges to **ceh.charges**.
3. update the manuals.


thfroitzheim (approved):
Looks good

PR #1020, opened 2024-04-24
error tblite/ptb + solvation

Albkat:


src/prog/main.F90
```diff
@@ -224,6 +224,14 @@ subroutine xtbMain(env, argParser)
       call parseArguments(env, argParser, xcontrol, fnv, lgrad, &
          & restart, gsolvstate, strict, copycontrol, coffee, printTopo, oniom, dipro, tblite)
 
+      
+      ! TEMPORARY: no solvation available for PTB and tblite !
+      if (set%mode_extrun == p_ext_tblite .or. set%mode_extrun == p_ext_ptb) then
+         if (allocated(set%solvInput%solvent)) then
+            call env%error("Solvation is not implemented for PTB/tblite", source)
```
MtoLStoN, line 231: As this is in main, this could be a direct env%terminate instead of a raised error that will terminate at the next checkpoint.

Albkat: That is a good suggestion.
I would still merge this as it is, because the `xtbMain`  will be refactored next week.

PR #1060, opened 2024-06-24
Detect ifx (intel-llvm) when determining compiler settings

TyBalduf:
* Use `fc_id` throughout meson.build, rather than repeatedly calling `fc.get_id()`. 
* Wherever we previously branched for `"intel"` (ifort compiler), also branch for `"intel-llvm"` (ifx compiler).


awvwgk (approved):
Thanks for sharing. Looks good to me

PR #1061, opened 2024-07-02
Fixes #1054 by ensuring thread savety of hessian calculations with GFN-FF

### LINKED ISSUE #1054: Hessian calculations with GFN-FF fail
**Describe the bug**
Attempting to run different types of Hessian calculations (`--hess`, `--bhess`, `--ohess`) in combination with GFN-FF leads to crashes, even with an optimized starting geometry.

**To Reproduce**

1. happens with input file `c2h4.xyz` in the attached archive [hess.zip](https://github.com/user-attachments/files/15903475/hess.zip)
2. start `xtb` (binary release 6.7.0 from GitHub) with any of the following options:
    ```
    xtb c2h4.xyz --hess --gfnff --verbose >stdout 2>stderr
    xtb c2h4.xyz --bhess --gfnff --verbose >stdout 2>stderr
    xtb c2h4.xyz --ohess --gfnff --verbose >stdout 2>stderr
    ```
3. output showing the failed calculations for all three variants in the archive attached

**Expected behaviour**
Successful Hessian calculations with GFN-FF.

**Additional context**
These problems do not appear to happen without the `--gfnff` flag (that is with GFN2-xTB).

Thomas3R:
The distance array neigh%distances was not allocated thread save and was removed from the neigh class.
Now the distances are calculated from the atom coordinates and the translation vectors.

### DIFF
--- a/src/gfnff/gfnff_eg.f90
+++ b/src/gfnff/gfnff_eg.f90
@@ -197,6 +197,7 @@ subroutine gfnff_eg(env,mol,pr,n,ichrg,at,xyz,sigma,g,etot,res_gff, &
    real(wp),allocatable :: xtmp(:)
    type(tb_timer) :: timer
    real(wp) :: dispthr, cnthr, repthr, hbthr1, hbthr2
+   real(wp) :: dist(n,n)
    real(wp) :: ds(3,3)
    real(wp) :: convF  !convergence factor alpha, aka ewald parameter 
    logical, allocatable :: considered_ABH(:,:,:)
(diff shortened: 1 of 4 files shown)


marcelmbn:
Given the assumption that the distance array cannot be allocated thread safe, this seems like a reasonable change to me.
However, it's principle-wise _not a good development to move away from modularized calculation of static variables (like a distance array) towards calculating it from scratch whenever needed_.
**Thus, do you think there's any chance to allocate it in a thread-safe way?**

Thomas3R:
Of course, a distance array could be used in a thread-saving way. However, I recommend implementing this together with a revision of the force field parallelization. This would be a major effort to implement either way.

marcelmbn (approved):
Again, assuming that there's no better solution using the `distances` array of the `neigh` type in a thread-safe manner, this seems like a reasonable change.

PR #1071, opened 2024-07-30
Add Hessian to C-api

TyBalduf:
Adds an API interface to the calculator type's `hessian` method. Unlike the underlying method, the parameters (e.g. step size and list of atoms to include) and other return values (dipole gradient) are all optional.

### DIFF
--- a/include/xtb.h
+++ b/include/xtb.h
@@ -239,6 +239,18 @@ xtb_singlepoint(xtb_TEnvironment /* env */,
                 xtb_TCalculator /* calc */,
                 xtb_TResults /* res */) XTB_API_SUFFIX__VERSION_2_0_0;
 
+/// Perform hessian calculation
+extern XTB_API_ENTRY void XTB_API_CALL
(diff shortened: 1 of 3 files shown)


TyBalduf:
The Mac cpx failure seems unrelated, as all the changes here should only affect the C-API.

TyBalduf:
@awvwgk sorry to ping you on this pull request as well, but I wasn't sure who else would be comfortable reviewing changes to the C-API.

marcelmbn:
Can you update your feature branch with the latest updates from `grimme-lab:main` and check if the CI failure is still present?

marcelmbn (dismissed):
LGTM.

marcelmbn:
After merging the first of your C-API changes, there seems to be a conflict. Please resolve, so that we can merge this PR as well.

marcelmbn (approved):

TyBalduf:
Was able to resolve the conflict, just needed to bump the API version and make sure that different test cases were affecting each others reference values. Should be good to go now.

PR #1072, opened 2024-07-31
Make xtb compilable with latest tblite commit (6b7ee4e) EDIT: (6426ee4)

marcelmbn:
- receive `error_type` from `tblite`'s `new_*_calculator` call
- more recent `mstore`  version for _f_-block compounds
- adapt PTB tests to changed ordering of _d_ orbitals in `tblite`

### DIFF
--- a/test/unit/test_ptb.F90
+++ b/test/unit/test_ptb.F90
@@ -215,7 +215,7 @@ contains
       call check_(error, ints%overlap(2, 3), overlap_exp(3), thr=thr)
       call check_(error, ints%overlap(1, 15), overlap_exp(4), thr=thr)
       call check_(error, ints%overlap(1, 23), overlap_exp(5), thr=thr)
-      call check_(error, ints%overlap(12, 22), overlap_exp(6), thr=thr)
+      call check_(error, ints%overlap(11, 22), overlap_exp(6), thr=thr)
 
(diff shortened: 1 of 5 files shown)


test/unit/test_ptb.F90
```diff
@@ -215,7 +215,7 @@ subroutine test_ptb_overlap(error)
       call check_(error, ints%overlap(2, 3), overlap_exp(3), thr=thr)
       call check_(error, ints%overlap(1, 15), overlap_exp(4), thr=thr)
       call check_(error, ints%overlap(1, 23), overlap_exp(5), thr=thr)
-      call check_(error, ints%overlap(12, 22), overlap_exp(6), thr=thr)
+      call check_(error, ints%overlap(11, 22), overlap_exp(6), thr=thr)
```
thfroitzheim, line 218: Maybe add a comment to make clear which pair of orbitals the number corresponds to or you can change to checking the complete matrix

marcelmbn: I agree that in the long-term, a comparison of the full matrices would be more consistent and definitely better practice. For now, I've introduced annotations on which matrix elements are actually compared.

thfroitzheim (approved):
looks good to me

Issue #649, opened 2022-06-16
Translational entropy calculations


dgarayr:
Checking thermochemical calculations from xTB, I find the way in which translational entropy is calculated a bit weird (although results are right). The expression in the thermodyn subroutine from thermo.f90 for **s_tr** involves a simplified expression including two hard-coded values named magic4 and magic5.
```fortran
s_tr=magic4*(5.0_wp*log10(t)+3.0_wp*log10(molmass))-magic5
```
There is no direct reference to the source of these values, although they can be traced to an equation presented by [Stewart and Tseng, J. Am. Chem. Soc. 1971, 93, 5, 1273](https://pubs.acs.org/doi/pdf/10.1021/ja00734a047), which was derived from the Sackur-Tetrode equation as presented in _S. Glasstone, “Theoretical Chemistry”, 1944_.  Following the original text, it is found that the simplified equation is reached by rearranging constants, changing the logarithms to base 10, and converting P to atm and R to cal mol<sup>-1</sup> K<sup>-1</sup>. Although the expression is correct and thus the obtained values are valid, it is quite confusing to encounter this quite old and non-referenced expression for the translational entropy when the rest of magnitudes in the subroutine, including the own translational partition function, are computed directly from physical constants using common expressions for statistical thermodynamics.

Moreover, the employed expression drops the pressure dependence of translational entropy, assuming **P = 1** to cancel the logarithm. I could not find the pressure reference state used by xTB for thermochemical calculations neither in the module nor in the documentation, so it might be stated more explicitly at some point, maybe even in the output together with the temperature. Going a bit beyond this, I think the ideal feature would be to allow the user to change the pressure reference state as it is done with temperature, like in other codes such as Gaussian or ADF.

PR #1077, opened 2024-08-06
Add explanation for translation entropy constants

TyBalduf:
Worked out the constants following the expressions for translation entropy given (among other places) [here](https://wiki.fysik.dtu.dk/ase/ase/thermochemistry/thermochemistry.html)


src/thermo.f90
```diff
@@ -106,8 +106,10 @@ subroutine thermodyn(iunit,A_rcm,B_rcm,C_rcm,avmom_si,linear,atom,sym,molmass, &
    real(wp),parameter :: H = 6.626176D-27 ! PLANCK'S CONSTANT IN ERG-SECONDS
    real(wp),parameter :: AK = 1.3807D-16  ! BOLTZMANN CONSTANT IN ERG/DEGREE
    real(wp),parameter :: conv3 = amutokg*1000 ! 1.6606d-24
-   real(wp),parameter :: magic4 = 2.2868d0 ! unknown
-   real(wp),parameter :: magic5 = 2.3135d0 ! unknown
+   real(wp),parameter :: magic4 = 2.2868d0 ! R*ln(10)/2
+   real(wp),parameter :: magic5 = 2.3135d0 ! R*(ln[(kb/P°)*(2pi * kB * amutokg/h)^(3/2)] + 5/2)
+                                           ! All terms in `magic5` in SI units (excluding R)
+                                           ! P° corresponds to 1 atm in Pascals.
```
marcelmbn, line 112: Thanks for sharing! If you have the correct derivation of the constants, could you implement them into the code to avoid using a fixed float point number?

I'd suggest using the constants in `src/mctc/mctc_constants.f90` and (if possibly necessary for other procedures as well) add the conversion parameters to `src/mctc/convert.f90` to avoid computing these quite extensive constants in the `thermo` module.

marcelmbn (changes requested):

src/thermo.f90
marcelmbn: (Obviously, please check thoroughly if the calculated constants match the currently present fixed numbers, before proceeding.)

marcelmbn:
Would fix also https://github.com/grimme-lab/xtb/issues/649.

TyBalduf:
So I rewrote the translation entropy equation and it gives approximately the same result (free energy differs by ~10^-5 kcal/mol)

If I work out the equivalent of magic4 and magic5, they come out to:
* magic4 = 2.2879 (previously 2.28*69*)
* magic5 = 2.3149 (previously 2.31*35*)

As best I can tell, these must have been originally been worked out with slightly different definitions of the constants. Not all the constants used in xTB correspond to their exact current value. For example in thermo, these values are defined:
```
   real(wp),parameter :: R = 1.98726D0    ! GAS CONSTANT IN CALORIES/(MOLE*K) (should be 1.9872*04*)
```

Not sure if this is an acceptable difference and tests should be updated or this is too large a change and we just stick with the current magic constants, just providing some documentation on how they were likely calculated.

marcelmbn:
Thanks for the correct implementation of the constant!

1. Currently, the tests fail due to a ~10-5 value mismatch in the `thermo` unit tests, which is not very surprising.
2. I've checked your point regarding the gas constant and came to the same result. The right way to do it would probably be to update all necessary constants (if necessary) to their most recent SotA value and then update the reference unit test results with the updated values. IMHO, there's no reason to stick with old fixed numbers based on slightly wrong fundamental constants.
3. Finally, we should move all constants from `thermo.f90` into `mctc/mctc_constants.f90` and just `use` them at the beginning of `thermodyn`. Moreover, I'd also delete the old code, i.e. `magic4`, `magic5`, `s_tr_old`, ... as people could still look it up in the `git` history and we don't need it in the active code base.

Thank you in advance for your work on this project!

src/mctc/convert.f90
```diff
@@ -110,6 +120,7 @@ function m4bykgtoang4byamu() result(sitoang4byamu)
       real(wp) :: sitoang4byamu
       sitoang4byamu = (1.0e+10_wp)**4 / kgtoamu ! 16605390400000.0
    end function m4bykgtoang4byamu
+
```
marcelmbn, line 123:
```suggestion
```

marcelmbn:
@TyBalduf Just wanted to ping you again on this topic. We'd be very grateful if you find the time to update the related points in the code to the SotA standard you suggested, including the correct _ab initio_ constants, so that we can integrate everything into the `main` branch. 😊
If you are no longer working on it, a short note would also be nice.

TyBalduf:
I've been a little preoccupied of late and won't have a chance to work on this too soon. I can probably get to it eventually, its mainly a matter of updating tests. But I'm not actively working on it so if someone else wanted to take this on, there wouldn't be conflicting work.

thfroitzheim:
I updated the necessary constants from CODATA 2018 to be consistent with mctc-lib and calculated all necessary quantities directly from the fundamental constants. The resulting corrections change by about 10e-5 but the new values are just more accurate. Therefore, I adapted the tests to the new values. Should now be ready.

src/thermo.f90
```diff
@@ -292,9 +298,8 @@ subroutine oldthermo(a,b,c,avmom,linear,atom,sym,molmass,vibs,nvibs,escf, &
    real(wp),parameter :: conv1 = rcmtoau*autokcal*0.5_wp ! 1.4295718d-3
    real(wp),parameter :: conv2 = rcmtoau*autokJ*1000 ! 0.01196266D+3
    real(wp),parameter :: conv3 = amutokg*1000 ! 1.6606d-24
-   real(wp),parameter :: conv4 = 2.2868d0 ! unknown
-   real(wp),parameter :: conv5 = 2.3135d0 ! unknown
-   real(wp),parameter :: caltoj = autokj/autokcal
+   real(wp),parameter :: conv4 = 2.2868d0 ! approx R*ln(10)/2
+   real(wp),parameter :: conv5 = 2.3135d0 ! approx R*(ln[(kb/P°)*(2pi * kB * amutokg/h)^(3/2)] + 5/2)
```
marcelmbn, line 302: I'd suggest deleting these lines as they should not be needed anymore.

src/thermo.f90
```diff
@@ -222,10 +217,15 @@ subroutine thermodyn(iunit,A_rcm,B_rcm,C_rcm,avmom_si,linear,atom,sym,molmass, &
...
-   s_tr_old=magic4*(5.0_wp*log10(t)+3.0_wp*log10(molmass))-magic5
-   s_tr=R*((2.5_wp*log(t*kB)&
+   s_tr=R*((2.5_wp*log(T*kB)&
   &        +1.5_wp*log(amutoau*molmass/(twopi))&
   &        -log(atmtoau) + 2.5_wp))
+   ! Alternative form: 
+   ! s_tr_old=magic4*(5.0_wp*log10(t)+3.0_wp*log10(molmass))-magic5
+   ! with:
+   ! magic4 = R*ln(10)/2 approx 2.2868d0 
+   ! magic5 = R*(ln[(kb/P°)*(2pi * kB * amutokg/h)^(3/2)] + 5/2) approx 2.3135d0
```
marcelmbn, line 227: We could also delete these old legacy parts but might also be helpful for some point in the future where it possibly helps somebody in understanding the equations? I'd leave it up to you if we keep it or not.

src/mctc/convert.f90
```diff
@@ -57,6 +57,8 @@ module xtb_mctc_convert
 !  femtosectons to atomic time units
    real(wp), public, parameter :: fstoau = 41.3413733365614_wp
    real(wp), public, parameter :: autofs = 1.0_wp / fstoau
```
marcelmbn, line 59: Shouldn't we replace these two lines with dependencies on the `atomic_unit_of_time`? Both represent a relation between units of time (SI vs atomic units)?

marcelmbn: (Or the other way around would obviously also be possible)

marcelmbn (changes requested):
Looks good to me besides the few questions regarding constant relationships.

src/thermo.f90
thfroitzheim: I would keep this to explain the formula above. It is not directly obvious and long-term we need to clean up the thermo routine anyway. Then it is nice to have as much information as possible.

src/thermo.f90
thfroitzheim: Then we have to remove the full routine, which I did. I think it mainly calculates everything in wavenumbers and was never used in the code.

src/mctc/convert.f90
thfroitzheim: We can do this. I used the new value which comes from mctc-lib (our general goal should be to consistently use only mctc-lib values for all constants, as this will replace the internal mctc routines in xtb).

Issue #1079, opened 2024-08-07
Dock module not working in xtb v6.7.1


icamps:
**Describe the bug**
Running a simple dock calculation with two small molecules (CO2 and NO2) with `6.7.1 (edcfbbe)` and `6.7.1 (ec4f388) (bleed)`.

**To Reproduce**
Steps to reproduce the behaviour:

1. opt_CO2.sdf and opt_NO2.sdf: 
[opt_CO2.sdf.txt](https://github.com/user-attachments/files/16533136/opt_CO2.sdf.txt)
[opt_NO2.sdf.txt](https://github.com/user-attachments/files/16533137/opt_NO2.sdf.txt)

3. start `xtb` with: `dock opt_CO2.sdf opt_NO2.sdf --opt normal --etemp 298 --verbose`
4. run `xtb` with your options and the `--verbose` flag
5. output showing the error: 
[log-6.7.1.txt](https://github.com/user-attachments/files/16533139/log-6.7.1.txt)
[log-bleed.txt](https://github.com/user-attachments/files/16533140/log-bleed.txt)

Please provide all input and output file such that we confirm your report.

**Expected behaviour**
The docking calculation run ok.

**Additional context**

cplett:
Hi,
Thank you very much for reporting the bug. It appears to be a similar issue to #1078.
With the latest pull request #1089, the docking should work properly again.
In the meantime, I suggest using xtb version 6.7.0 or a compiled binary containing the latest pull request to use the docking.

marcelmbn:
Should be resolved by https://github.com/grimme-lab/xtb/pull/1089.

Issue #1078, opened 2024-08-06
Segfaults in aISS submodule with precompiled release 6.7.1


b-rousse:
**Describe the bug**
The aISS docking submodule segfaults in the latest xtb release (6.7.1) above a certain system size. I can confirm it is not related to OMP environment variables nor the ulimit -s unlimited command. Reverting to version 6.7.0 fixed the problem.

**To Reproduce**
Steps to reproduce the behaviour:
Use latest release 6.7.1, try: 
xtb dock PFOA.xyz cluster_min_3layer.xyz 

This is with release version 6.7.1.

Please note I cannot upload .xyz or .inp files, so files minimal_1.xyz and PFOA.xyz have been provided in .txt format.
[cluster_min_3layer.txt](https://github.com/user-attachments/files/16516354/cluster_min_3layer.txt)
[PFOA.txt](https://github.com/user-attachments/files/16516267/PFOA.txt)

**Additional notes**
I tried this on a node with 256 GB RAM, set OMP_NUM_THREADS=1, OMP_STACKSIZE=256G, and ulimit -s unlimited. It still crashed.

cplett:
Hi,
Thank you very much for reporting the bug. 
The pull request #1089 should fix it, and with the next release, the aISS should work properly again.
In the meantime, I suggest using xtb version 6.7.0 or a compiled binary containing the latest pull request to use the docking.

PR #1089, opened 2024-08-19
aISS update

cplett:
- Bug fixes for the aISS docking module related to #1078 and #1079 
- Ensuring that the correct charges and coordination numbers are used for the GFN1-xTB and GFN2-xTB calculations during the aISS docking algorithm

### DIFF
--- a/src/docking/search_nci.f90
+++ b/src/docking/search_nci.f90
@@ -1136,7 +1136,6 @@ contains
       if (mol%npbc > 0) then
          chk%wfn%q = real(set%ichrg, wp)/real(mol%n, wp)
       else
-              write(*,*) mol%z
             if (set%guess_charges == p_guess_gasteiger) then
                call iniqcn(mol%n, mol%at, mol%z, mol%xyz, set%ichrg, 1.0_wp, chk%wfn%q, cn, set%gfn_method, .true.)
             else if (set%guess_charges == p_guess_goedecker) then
--- a/src/docking/search_nci.f90
+++ b/src/docking/search_nci.f90
@@ -48,13 +48,14 @@ module xtb_docking_search_nci
    & gff_print, gfnff_param_dealloc
    use xtb_constrain_param, only: read_userdata
    use xtb_fixparam
-   use xtb_disp_ncoord, only: ncoord_gfn, ncoord_erf
+   use xtb_disp_ncoord, only: ncoord_gfn, ncoord_erf, ncoord_d3
    use xtb_scc_core, only: iniqshell
    use xtb_eeq, only: goedecker_chrgeq
    use xtb_basis, only: newBasisset
    use xtb_gfnff_neighbor, only: TNeigh
    use xtb_io_writer, only : writeMolecule
    use xtb_mctc_filetypes, only : generateFileName
+   use xtb_iniq, only: iniqcn
    implicit none
 
    private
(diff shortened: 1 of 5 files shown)


src/docking/search_nci.f90
```diff
@@ -1065,11 +1066,19 @@ subroutine restart_gff(env, mol, calc)
       call gfnff_param_dealloc(calc%topo)
       call newD3Model(calc%topo%dispm, mol%n, mol%at)
       call gfnff_set_param(mol%n, calc%gen, calc%param)
+      if (allocated(calc%neigh%nb)) deallocate(calc%neigh%nb)
       if (.not. allocated(calc%neigh%nb)) allocate (calc%neigh%nb(calc%neigh%numnb, mol%n, calc%neigh%numctr), source=0)
```
marcelmbn, line 1070: After the enforced deallocation in line 1069, `if (.not. allocated(calc%neigh%nb))` is not necessary again.

src/docking/search_nci.f90
```diff
@@ -1065,11 +1066,19 @@ subroutine restart_gff(env, mol, calc)
...
+      if (allocated(calc%topo%qfrag)) deallocate(calc%topo%qfrag)
       if (.not.allocated(calc%topo%qfrag)) &
```
marcelmbn, line 1072: Same as above.

src/docking/search_nci.f90
```diff
@@ -1116,15 +1125,30 @@ subroutine restart_xTB(env, mol, chk, calc, basisset)
       calc%maxiter = set%maxscciter
 
       call chk%wfn%allocate(mol%n, calc%basis%nshell, calc%basis%nao)
+      ! Make sure number of electrons is initialized an multiplicity is consistent
+      chk%wfn%nel = nint(sum(mol%z) - mol%chrg)
+      chk%wfn%nopen = mol%uhf
+      if (chk%wfn%nopen == 0 .and. mod(chk%wfn%nel, 2) /= 0) chk%wfn%nopen = 1
```
marcelmbn, line 1131: What about molecules, for which `mol%uhf` has already been assigned to, e.g. `2`? Wouldn't this be (incorrectly) determined to `0`?

src/iff/iff_prepare.f90
```diff
@@ -193,7 +193,7 @@ subroutine precomp(env, iff_data, mol, etot, mol_num)
       !> the SP
       call singlepoint &
          &       (env, mol, chk, calc, egap, set%etemp, set%maxscciter, 2,&
-         &        exist, lgrad, acc, etot, g, sigma, res)
+         &        .false., lgrad, acc, etot, g, sigma, res)
```
marcelmbn, line 196: Can you shortly explain why this is now fixed?

marcelmbn (commented):
Please check the remaining questions. @cplett

src/docking/search_nci.f90
cplett: It is removed now.

src/docking/search_nci.f90
cplett: Should not be the case. `chk%wfn%nopen` is set once to `mol%uhf` in line 1130 and only changed in line 1131 if `mol%uhf` is zero and the number of electrons is odd. This is similar to the `xtb` main routine:
https://github.com/grimme-lab/xtb/blob/043d587270844d953b0f5c1fe88e34a0eccb3b58/src/prog/main.F90#L600-L603

src/iff/iff_prepare.f90
cplett: The boolean `exist` was never assigned a value. However, the `singlepoint` subroutine expects the boolean with `intent(in)`. Even though this was not yet causing a program crash, it might have led to memory problems in the future.

src/docking/search_nci.f90
cplett: Also removed now.

marcelmbn (approved):

PR #1097, opened 2024-09-12
GFN-FF reparameterization for lanthanides and extension to actinides

Thomas3R:
This version enables GFN-FF calculations for systems including elements up to lawrencium (103).
The performance of the method has been improved for lanthanides.
Furthermore, the neighbor list can be manually adjusted via the detailed input file (this feature is not available with PBC).


src/gfnff/gfnff_ini.f90
```diff
@@ -2184,6 +2199,79 @@ integer function itabrow6(i)
...
+        nnb = neigh%nb(neigh%numnb, i, iTr)
+        idx=0
+        do count_idx=1, nnb
+          idx = idx + 1
+          inb=neigh%nb(idx,i,iTr) ! atom index of neighbor of i
+          if (inb.eq.0) cycle
+          ! check if neighbor is H
+          if (mol%at(inb).eq.1) then
+            ! remove hydrogen as neighbor if charge is gt threshold
+            if (topo%qa(inb).gt.-0.0282) then
```
marcelmbn, line 2227:
```suggestion
            if (topo%qa(inb).gt.-0.0282_wp) then
```

marcelmbn: Shouldn't we put this number to the top and mark it as a parameter?

marcelmbn:
General comment: We should have tests that cover the new functionality and new parameters.

marcelmbn:
Please fix the `forrtl: severe (174): SIGSEGV, segmentation fault occurred` that seems to occur reproducibly in the `pbc` tests of `gfnff` before we can merge.

PR #1103, opened 2024-09-17
GFN-FF reparameterization for lanthanides and extension to actinides

Thomas3R:
This version enables GFN-FF calculations for systems including elements up to lawrencium (103).
The performance of the method has been improved for lanthanides.
Furthermore, the neighbor list can be manually adjusted via the detailed input file (this feature is not available with PBC).
Unit tests have been added to test Ln/An energies, gradients, and the Ln/An-H neighbor list setup.


src/gfnff/gdisp0.f90
```diff
@@ -130,6 +130,7 @@ subroutine weight_references_d4(dispm, nat, atoms, wf, cn, gwvec, gwdcn)
          norm = norm + gw
          dnorm = dnorm + 2*wf*(dispm%cn(iref, ati) - cn(iat)) * gw
       end do
+      if (norm.eq.0.0_wp) cycle
```
marcelmbn, line 133: What does this `cycle` lead to? Why has it not been required before?

src/setparam.f90
```diff
@@ -516,6 +516,10 @@ module xtb_setparam
 !! ------------------------------------------------------------------------
    !> PTB settings
    type(TPTBSetup) :: ptbsetup
+   !> GFN-FF manual setup of nb list via xcontrol
+   !  allows a maximum of 164 atoms neighbors to be changed
+   !  ffnb(42,i) stores the number of neighbors of atom i
+   integer :: ffnb(42,164) = -1
```
marcelmbn, line 522: Do you see any chance to make the size of this array dependent on the input (by scanning the input first and then allocating the required memory)? If not, this is still a valid approach, but we should at least document why it's not reasonable and we take the approach of a fixed memory variable.

src/set_module.f90
```diff
@@ -1499,6 +1500,66 @@ subroutine set_gfn(env,key,val)
...
+           nonb = .true.
+         endif
+         ! set rest of ffnb to 0
+         set%ffnb(k+1:41, i_ffnb) = 0
+      endif
+   enddo
+   if (nonb) then
+     set%ffnb(2:,i_ffnb) = 0
+   else
+     set%ffnb(42,i_ffnb) = k - 1 ! number of neighbors of atom set%ffnb(1,i_ffnb)
```
marcelmbn, line 1558: Does this line mean that maximally 42 neigbours are allowed, or in other words, that the array always contains 42 neighbours and usually, most of them being zero?

marcelmbn (commented):
Thank you for your contribution, and extending it with some tests for the new parameterization and all associated functions!
Except for the few questions, this should be good to go as we have already had a look on the first pull request for the implementation in https://github.com/grimme-lab/xtb/pull/1097.

src/gfnff/gdisp0.f90
Thomas3R: It's something I found while debugging with valgrind. I would expect that it has been required before, but the compiler and compiler options we use already seem to handle problems with division by zero. However, we should not divide by zero, that is why I added the cycle to continue with the next atom if the norm is zero for the current atom. As checked by the unit tests, this does not influence energy, gradient, or sigma.

src/setparam.f90
marcelmbn: Let's shift it to another PR, @Thomas3R.

src/set_module.f90
marcelmbn: Resolved by in-person discussion.

marcelmbn (approved):

Thomas3R:
Thanks for the nice review @marcelmbn

PR #1105, opened 2024-09-18
Allocate set%ffnb dynamically.

Thomas3R:
The array used for the adjustment of the GFN-FF neighbor list (set%ffnb) is now allocated dynamically.
The first dimension is always the maximum number of neighbors allowed in GFN-FF (42 since 6.7.1) and the second dimension is the number of atoms that the neighbor list should be adjusted for.

### DIFF
--- a/src/gfnff/gfnff_ini.f90
+++ b/src/gfnff/gfnff_ini.f90
@@ -2450,10 +2450,12 @@ subroutine gfnff_topo_changes(env, neigh)
    integer :: int_tmp(40)
    integer :: i,j,idx,iTr,d1,d2,numnb,nnb
 
-   ! check if hardcoded size of ffnb is still up to date
-   if (size(set%ffnb, dim=1).ne.neigh%numnb) call env%error('The array set%ffnb has not been adjusted to changes in neigh%numnb.', source)
    ! only do something if there are changes stored in set%ffnb
-   if(set%ffnb(1,1).ne.-1) then
+   if(allocated(set%ffnb)) then
+      ! check if hardcoded size of ffnb is still up to date
+      if (size(set%ffnb, dim=1).ne.neigh%numnb) call env%error('The array set%ffnb has not been adjusted to changes in neigh%numnb.', source)
+      ! there should not be any "-1" in set%ffnb anymore, if it was set up correctly
+      if (any(set%ffnb.eq.-1)) call env%error('GFN-FF neighbor list could not be adjusted!', source)
       d2=size(set%ffnb, dim=2)
       do i=1, d2
          if (set%ffnb(1,i).eq.-1) exit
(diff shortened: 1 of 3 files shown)


src/set_module.f90
```diff
@@ -1500,6 +1503,51 @@ subroutine set_gfn(env,key,val)
...
+! and allocate set%ffnb accordingly
+subroutine alloc_ffnb(env, fname)
+   type(TEnvironment), intent(inout) :: env
+   character(len=*),intent(in)  :: fname
+   character(len=*), parameter :: source = 'alloc_ffnb'
+   character(len=:),allocatable :: line
+   integer :: copy, err
+   integer :: id, ie
+   logical :: is_ffnb_block
+   ! character,private,parameter :: flag = '$' ! is defined for this module
```
marcelmbn, line 1516:
```suggestion
   integer :: copy, err
   integer :: id, ie
   logical :: is_ffnb_block
```

src/set_module.f90
```diff
@@ -1500,6 +1503,51 @@ subroutine set_gfn(env,key,val)
...
+   integer :: n_changes ! number of atoms that neigh%nb should be adjusted for
```
marcelmbn, line 1517:
```suggestion
   integer :: copy, err
   integer :: id, ie, n_changes ! n_changes: number of atoms that neigh%nb should be adjusted for
   logical :: is_ffnb_block
```

src/set_module.f90
```diff
@@ -1500,6 +1503,51 @@ subroutine set_gfn(env,key,val)
...
+
+   copy=-1 ! do not copy the control file
+   n_changes = 0
+   is_ffnb_block = .false.
```
marcelmbn, line 1521:
```suggestion

   integer :: id, ie
   logical :: is_ffnb_block = .false.
   integer :: copy = -1
   integer :: n_changes = 0 ! number of atoms that neigh%nb should be adjusted for
```

src/gfnff/gfnff_ini.f90
```diff
@@ -2450,10 +2450,12 @@ subroutine gfnff_topo_changes(env, neigh)
...
 
-   ! check if hardcoded size of ffnb is still up to date
-   if (size(set%ffnb, dim=1).ne.neigh%numnb) call env%error('The array set%ffnb has not been adjusted to changes in neigh%numnb.', source)
    ! only do something if there are changes stored in set%ffnb
-   if(set%ffnb(1,1).ne.-1) then
+   if(allocated(set%ffnb)) then
+      ! check if hardcoded size of ffnb is still up to date
+      if (size(set%ffnb, dim=1).ne.neigh%numnb) call env%error('The array set%ffnb has not been adjusted to changes in neigh%numnb.', source)
+      ! there should not be any "-1" in set%ffnb anymore, if it was set up correctly
+      if (any(set%ffnb.eq.-1)) call env%error('GFN-FF neighbor list could not be adjusted!', source)
```
marcelmbn, line 2458: Just because in `set_ffnb` (if it was correctly executed) all numbers are positive integers and this is some sort of check that everything worked fine?

marcelmbn (commented):
Just commit the "style" suggestion and resolve the single question, then it should be good to go!

src/gfnff/gfnff_ini.f90
Thomas3R: Exactly, `set%ffnb` is initialized to -1, and all entries should be updated with values greater or equal to zero.

marcelmbn (approved):
LGTM.

PR #1140, opened 2024-11-27
Fix optimization when davidson fails

foxtran:
In rare cases when davidson fails, solver_sspevx was called. Unfortunately, it did not update fail variable, so, if it even was successful, fail variable still was true and therefore optimization was failing.

This patch fixes this issue.


Albkat (approved):
Thanks for the patch, lgtm

PR #1153, opened 2025-01-09
CEH gradients

Albkat:
The implementation of the numerical charge gradients `num_grad_chrg` (for CEH).
The output is written to **ceh.grad**.
@lukaswittmann, please be careful with the [output](https://github.com/grimme-lab/xtb/blob/1186b5a4b3338a78e6ebe4e95d7dec924c536e90/src/tblite/calculator.F90#L495-L497) format.

### DIFF
--- a/src/optimizer.f90
+++ b/src/optimizer.f90
@@ -1007,6 +1007,7 @@ pure subroutine solver_ssyevx(n,thr,A,U,e,fail)
    j=1
    call lapack_syevx('V','I','U',n,A,n,dum,dum,j,j,thr, &
    &           i,e,U,n,work,lwork,iwork,ifail,info)
+   fail = .false.
    if (info.ne.0) fail = .true.
 
    deallocate(iwork,work,ifail)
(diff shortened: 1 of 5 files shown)


Albkat:
The CI failure is due to the missing LAPACK/BLAS libraries in Ubuntu 24.04.1.  

It seems to work with Ubuntu 24.04.5 ([run](https://github.com/Albkat/xtb/actions/runs/12691275702/job/35373965082))
As the updated runners are still being [rolled out](https://github.com/actions/runner-images/issues/10636), the potential fix might be completed around the end of January.

lukaswittmann (dismissed):
Thank you for your great work, looks good to me.

lukaswittmann:
I renamed the file where the ceh numerical gradients are written to for consistency reason.

Albkat:
I added the CLI argument description:
```
--ceh grad [REAL]
```
to calculate numerical gradients (with an optional step size specification).

Issue #1147, opened 2024-12-12
Spelling of furan


jevandezande:
Solvent input only accepts `furane`, which is a much less common spelling of furan.

Ideally it would use the common English name, or accept both.

TyBalduf:
Related issues:

* https://github.com/grimme-lab/xtb/issues/1127
* https://github.com/grimme-lab/CPCM-X/issues/42

TyBalduf:
In general, I think it would be beneficial for xTB to adopt a solvent normalization scheme similar to what is done in [CPCM-X](https://github.com/grimme-lab/CPCM-X/blob/ca32f0f3b27b4f481882b07531a62c6844cadd07/src/cpcmx/data.f90#L68). It would presumably just be a small update to [normalizeSolventName](https://github.com/grimme-lab/xtb/blob/c3cfd38d6753f92a07a916a47a52114af9e6b989/src/solv/model.f90#L216), which currently just handles lowercasing the name.

 Currently, any inconsistency in the names used for solvents (e.g. xTB only accepts `DCM` or `ch2cl2` while CPCM-X until recently only accepted `methylenechloride`) will prevent CPCM_X from being used (or any future subprogram with differences in solvent name.

PR #1160, opened 2025-01-27
Solvent aliases

TyBalduf:
Adds additional solvent aliases for various supported solvents.

`normalizeSolventName` now chooses a consistent name to use for a given solvent internally.
 
Addresses https://github.com/grimme-lab/xtb/issues/1147


lukaswittmann (approved):
Looks good to me.

PR #1166, opened 2025-02-01
Fix double print of ancopt header

### LINKED ISSUE #1144: Ancopt header printed twice
**Describe the bug**
The header for Ancopt is printed twice: once in [main.F90](https://github.com/grimme-lab/xtb/blob/99b35832c0bc16e6b6cdb12cd1a55f8da1a80216/src/prog/main.F90#L825) and once in [optimizer.F90](https://github.com/grimme-lab/xtb/blob/99b35832c0bc16e6b6cdb12cd1a55f8da1a80216/src/optimizer.f90#L307).

```
 ----------------------------------------------------------- 
     |                   =====================                   |
     |                        A N C O P T                        |
     |                   =====================                   |
     |               Approximate Normal Coordinate               |
     |                Rational Function Optimizer                |
      ----------------------------------------------------------- 
      ----------------------------------------------------------- 
     |                   =====================                   |
     |                        A N C O P T                        |
     |                   =====================                   |
     |               Approximate Normal Coordinate               |
     |                Rational Function Optimizer                |
      ----------------------------------------------------------- 
```

**To Reproduce**
Run a geometry optimization with the default Ancopt optimizer. The header will be printed twice

**Expected behaviour**
Only one header

foxtran:
Fixes #1144

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -821,9 +821,6 @@ contains
             deallocate (set%opt_engine)
             call set_opt(env, 'engine', 'pbc_lbfgs')  ! use lbfgs
          end if
-         if (set%opt_engine == p_engine_rf) &
-            call ancopt_header(env%unit, set%veryverbose)
-         !! Print ANCopt header
 
          ! start optimization timer !
          call start_timing(3)


TyBalduf (approved):
Looks good to me. I think it makes sense removing the header here and having the geopt module handle writing it's own header.

TyBalduf (approved):

marvinfriede (approved):

PR #1167, opened 2025-02-01
Fix multiple open of xtbscreen.xyz in local.f90 (xtb/iff test)

foxtran:
By some reason, at https://github.com/grimme-lab/xtb/blob/c3cfd38d6753f92a07a916a47a52114af9e6b989/src/local.f90#L278 `xtbscreen.xyz` was opened at first scope level, but the file was closed at second scope level, and therefore, if the following check was failed:
https://github.com/grimme-lab/xtb/blob/c3cfd38d6753f92a07a916a47a52114af9e6b989/src/local.f90#L360
the close statement does not work for `xtbscreen.xyz` and then, at the next iteration, this file was opened at the second time.

### DIFF
--- a/src/local.f90
+++ b/src/local.f90
@@ -494,14 +494,12 @@ subroutine local(nat,at,nbf,nao,ihomo,xyz,z,focc,s,p,cmo,eig,q,etot,gbsa,basis,r
          enddo
       enddo
       new=k
-
-      if(set%pr_local) then 
-      call close_file(iscreen)
-      end if
       deallocate(wbo)
 
-
    endif
+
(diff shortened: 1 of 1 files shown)


TyBalduf (approved):
Do you have an example of an input where this error was occurring and that it's now resolved? Or I guess how did you find this?

The fix looks good/reasonable, I'm just curious if it's a rare case to hit or if it happens fairly regularly and has just gone unnoticed. I know some platforms/compilers are a bit more permissive about having multiple open handles to a file.

foxtran:
I used `xtb/iff` test from CMake build system for finding this.

marvinfriede (approved):

PR #1168, opened 2025-02-01
Fix exit condition in calc_distances

### LINKED ISSUE #621: NaN results in H2 geometry optimisation
**Describe the bug**
I'm writing some code that's doing automated exploration of chemical reaction space, using GFN2-xTB to optimise individual product molecules. On occasion, this involves running geometry optimisations on H<sub>2</sub>. Usually this is fine, but on maybe 1% of these occasions, xTB starts throwing NaN values all throughout it's output. I've not observed this with any other molecules running with the same settings, only H<sub>2</sub>, and the vast majority of the time the calculation runs fine. 

The issue appears to stem from the exact geometry used, as a fresh H<sub>2</sub> geometry with a bond length of 0.7 Ang will work normally while a specific geometry that has failed before will continue to fail on reruns. Furthermore, despite outputting mostly NaNs, the faulty calculation believes it has run successfully, such that it doesn't trigger any errors in my code until it's too late.

**To Reproduce**

1. happens with input `xtbin.xyz` (doesn't happen with input `h2.xyz`)
2. run `xtb --iterations 1000 --opt normal --grad xtbin.xyz` 
3. observe output past the Final Singlepoint stage for point of failure

**Inputs**
Sorry for not uploading, Github won't take xyzs.

`xtbin.xyz`:
```
    2
  0.000000000000000E+000
H      -2.76516815      1.55822385    -15.39243995
H      -2.75749951      0.78983182    -15.27941415
```

`h2.xyz`:
```
2

H 0.7 0.0 0.0
H 0.0 0.0 0.0
```

**Output**
See [xtb.log](https://github.com/grimme-lab/xtb/files/8621965/xtb.log) past the Final Singlepoint stage.

**Expected behaviour**
Calculation is expected to run as it would when the same command is run with `h2.xyz` (rough geometry created by hand) instead of `xtbin.xyz` (geometry created by code).

### LINKED ISSUE #850: Invalid characters during H2 optimization.
Hi
I have only seen this bug during the optimization of Hydrogen molecule, in which some invalid characters appear at test.out file.

**To Reproduce**
xtb test.xyz --verbose --opt > test.out

I have attached all files, input and output.
[test-H2.zip](https://github.com/grimme-lab/xtb/files/12384290/test-H2.zip)

Regards...

foxtran:
Probably affects #621

Before patch, one distance were skipped between atoms `# N` and `# N-1`. For single-bonded systems (like H2) it is critical since it means that they were not bounded and then a lot of artefacts happen. 

Fixes #850. Changes in output:
```diff
  * 1 selected distances

      #   Z          #   Z                                           value/Å
-     0   0      0   0                                     0.0000000 (max)
+     1   1 H        2   1 H                                       0.7769870 (max)

- * 0 distinct bonds (by element types)
+ * 1 distinct bonds (by element types)

    Z      Z             #   av. dist./Å        max./Å        min./Å
+   1 H    1 H           1     0.7769870     0.7769870     0.7769870
```

### DIFF
--- a/src/main/geometry.f90
+++ b/src/main/geometry.f90
@@ -381,8 +381,8 @@ subroutine calc_distances(n,at,xyz,bond,maxdist,ndist,dist,id, &
          jat = at(j)
          r  = sqrt(sum((xyz(:,i)-xyz(:,j))**2))
          if (bond(j,i).gt.0) then
-            m = m+1
             if (m.ge.maxdist) exit get_dist
+            m = m+1
             ndel(iat,jat) = ndel(iat,jat)+1
             maxdel(iat,jat) = max(maxdel(iat,jat),r)
             mindel(iat,jat) = min(mindel(iat,jat),r)


TyBalduf:
Might be missing something, but I don't see how this would affect https://github.com/grimme-lab/xtb/issues/621. This function seems to only be used for printing a summary of atom distances at the end of a calculation, but the NaN's in the linked issue start to occur during the initial SCF.

The PR here does fix a real bug, as the outputs from the linked issue do show that the "selected distances" section isn't being printed, which should only occur if no distances were found. More generally, the summary should have been printing N(N-1)/2 distances, but it presumably has always printed one less distance.

The code change looks reasonable to me, but I don't think it resolves https://github.com/grimme-lab/xtb/issues/621.

TyBalduf (approved):

foxtran:
@TyBalduf, yep. I found more bugs with H2 system, so it may not fix issue with #621

foxtran:
@Albkat, what is about this PR?

Albkat (approved):

PR #1170, opened 2025-02-01
Fix division by zero at geoopt

### LINKED ISSUE #500: GFN-FF optimization fails with NaN gradient norm
GFN-FF optimization of a simple organometallic complex fails with NaN gradient norm. This seems to be a bit random as running the calculation sometimes works and sometimes fails.
 
[test.zip](https://github.com/grimme-lab/xtb/files/6786652/test.zip)

foxtran:
Fixes #500

### DIFF
--- a/src/constr.f90
+++ b/src/constr.f90
@@ -613,7 +613,11 @@ Subroutine dphidrPBC(mode,nat,xyz,i,j,k,l,vTrR,vTrB,vTrC,phi,&
       dphidrj=0
       dphidrk=0
       dphidrl=0
-      onenner=1.0d0/(nan*nbn)
+      if (abs(nan*nbn).gt.eps) then
+         onenner=1.0d0/(nan*nbn)
+      else
+         onenner=0.0d0
+      endif
    else
       onenner=1.d0/nenner
    endif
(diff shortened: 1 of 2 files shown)


src/constr.f90
```diff
@@ -613,7 +613,11 @@ Subroutine dphidrPBC(mode,nat,xyz,i,j,k,l,vTrR,vTrB,vTrC,phi,&
       dphidrj=0
       dphidrk=0
       dphidrl=0
-      onenner=1.0d0/(nan*nbn)
+      if (abs(nan*nbn).gt.eps) then
```
TyBalduf, line 616: It might be worth clarifying what the purpose of this check is. It seems like we would only reach this condition if for some reason the magnitude of one or both of these surface normals was very small. The earlier check already determines if the normals are nearly parallel.

Is there a specific example you have found that the code change here fixes?

src/optimizer.f90
```diff
@@ -850,7 +850,7 @@ subroutine relax(env,iter,mol,anc,restart,maxcycle,maxdispl,ethr,gthr, &
       write(env%unit,'(5x,"change   ",e18.7,1x,"Eh")')                      echng
       write(env%unit,'(3x,"gradient norm :",f14.7,1x,"Eh/α")',advance='no') gnorm
       write(env%unit,'(3x,"predicted",e18.7)',advance='no')                 depred
-      write(env%unit,'(1x,"("f7.2"%)")')         (depred-echng)/echng*100
+      write(env%unit,'(1x,"("f7.2"%)")')         (depred-echng)/(echng+1e-34_wp)*100
```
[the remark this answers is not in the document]
foxtran: I found since I am using the same technique as for other issues. :-)

src/constr.f90
foxtran: >  The earlier check already determines if the normals are nearly parallel.

You can be here if `nan` or `nbn` equal zero and then you divide by zero :) 

> Is there a specific example you have found that the code change here fixes?

Yep, you can detect this with input provided in #500.

TyBalduf: I'm surprised the example from https://github.com/grimme-lab/xtb/issues/500 goes through this function. I assumed from the name it only applied for PBC structures, but the example from the issue is molecular.

foxtran: ¯\_(ツ)_/¯

foxtran:
@Albkat, can we merge it?

Albkat (approved):
LGTM

PR #1171, opened 2025-02-01
Fix L-BFGS memory accesses

foxtran:
This patch fixes some incorrect allocations to their proper size for using `anc` and `hdiag` variables in array statements like:

here for hdiag:
https://github.com/grimme-lab/xtb/blob/c3cfd38d6753f92a07a916a47a52114af9e6b989/src/relaxation_engine.f90#L849
(`d` and `q` have size `nvar`, while `hdiag` had size `nat3`)

here for anc:
https://github.com/grimme-lab/xtb/blob/c3cfd38d6753f92a07a916a47a52114af9e6b989/src/relaxation_engine.f90#L1053
(`displacement` has size `nvar`, while `anc` had size `nat3`)

### DIFF
--- a/src/relaxation_engine.f90
+++ b/src/relaxation_engine.f90
@@ -484,7 +484,6 @@ subroutine l_ancopt &
    real(sp), allocatable :: eig(:)
    real(wp), allocatable :: trafo(:,:)
    real(wp), allocatable :: hdiag(:)
-   real(wp), allocatable :: anc(:)
    real(wp), allocatable :: xyz0(:,:)
    type(convergence_log), allocatable :: avconv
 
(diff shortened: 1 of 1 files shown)


TyBalduf (commented):
Code change looks reasonable, though I do see a test failure for the one of Mac CI runs. Maybe it was just fluke and it will go away in a new CI run.

foxtran:
> Maybe it was just fluke and it will go away in a new CI run.

No, there is a some bug which I still do not catch  :(

foxtran:
@marvinfriede, what is about this PR?

marvinfriede (approved):

PR #1172, opened 2025-02-01
Accept any length symbols' strings

foxtran:
Noticed with xtb/vertical test, where length of symbols' strings equals one

### DIFF
--- a/src/mctc/symbols.f90
+++ b/src/mctc/symbols.f90
@@ -284,7 +284,7 @@ pure function findSymbol(list, symbol) result(position)
 
    position = 0
    do iSym = 1, size(list)
-      if (symbol == list(iSym)) then
+      if (trim(symbol) == trim(list(iSym))) then
          position = iSym
          exit
       end if
--- a/src/mctc/symbols.f90
+++ b/src/mctc/symbols.f90
@@ -330,6 +330,10 @@ pure subroutine appendSymbol(list, nList, symbol)
    !> Elements symbol
    character(len=*), intent(in) :: symbol
 
+   if (len(symbol) > len(list)) then
+     error stop "Programmer's bug: len(symbol) > len(list)"
+   end if
+
    if (nList >= size(list)) then
       call resize(list)
    end if
--- a/src/mctc/symbols.f90
+++ b/src/mctc/symbols.f90
@@ -246,7 +246,7 @@ subroutine getIdentitySymbol(nId, identity, symbol)
    integer, intent(out) :: nId
 
    !> Element symbols
-   character(len=symbolLength), intent(in) :: symbol(:)
+   character(len=*), intent(in) :: symbol(:)
 
    !> Chemical identity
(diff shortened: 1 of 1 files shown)


src/mctc/symbols.f90
```diff
@@ -246,7 +246,7 @@ subroutine getIdentitySymbol(nId, identity, symbol)
    integer, intent(out) :: nId
 
    !> Element symbols
-   character(len=symbolLength), intent(in) :: symbol(:)
+   character(len=*), intent(in) :: symbol(:)
```
TyBalduf, line 249: What problem does this solve? The array does accommodate symbols of length 1, it will just be padded out to `symbolLength` with spaces.

I also don't think an array of `character(len=*)` will work in general. All the strings in an array have to be the same length.

foxtran: > it will just be padded out to symbolLength with spaces.

No, it will not.

For example, in the following code it does not happen (except if your compiler clever enough, but then you just need to put functions into different translation units):
```fortran
program main
character(len=1), parameter :: sym = "K"
call print_sym(sym)
end program main

subroutine print_sym(sym)
character(len=*), intent(in) :: sym
integer, parameter :: unit = 6
call write_sym(unit, sym)
end subroutine print_sym

subroutine write_sym(unit, sym)
character(len=4), intent(in) :: sym
integer, intent(in) :: unit
write(6, *) sym
end subroutine write_sym
```

gfortran is not clever, then with it, you can just compile it with `-fcheck=all`:
```
gfortran test.f90 -fcheck=all
```
and then run it:
```
./a.out
```
and see the message:
```
At line 12 of file test.f90
Fortran runtime error: Actual string length is shorter than the declared one for dummy argument 'sym' (1/4)
```

Exactly the same stuff happens here and this PR resolves it :)

> All the strings in an array have to be the same length.

It is a general rule for any fortran-strings arrays

TyBalduf: I see, sorry I was thinking about it backwards as if this were the declaration for an actual array rather than the signature of the dummy argument.

The change makes sense to make handling of element symbols more flexible.

awvwgk (changes requested):
This would not work if you don't account for the variable character length in the allocation of the temporary array.

foxtran:
@awvwgk, why does not it work? `character(len=:), allocatable :: str(:)` can be also passed into `character(len=*), allocatable :: str(:)`.

See an example here: https://godbolt.org/z/fc7nTKxnP

awvwgk:
Because the allocation in the procedure still allocates using a fixed size, which would result in silent truncation. There is nothing against allocatable character arrays with allocatable length here (except spotty compiler Support, as usual).

foxtran:
@awvwgk, I have inserted a check that input string is no longer than `symbolLength`. Another way is to avoid known-length strings and use everywhere `character(len=:), allocatable, dimension(:)` strings arrays.

PR #1176, opened 2025-02-05
Fix isotope hessian

marcelmbn:
### Problem

- Isotope masses were overridden by physical values.

### Fix

- shift initialization of atomic masses directly to line directly after allocation of the array `atmass` (see `init_split`).

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -420,6 +420,8 @@ contains
       else
          call init_metadyn(mol%n, metaset%maxsave)
       end if
+      !> Initialize the atomic masses with the physical constants
+      atmass = atomic_mass(mol%at) * autoamu ! from splitparam.f90
       call load_rmsdbias(rmsdset, mol%n, mol%at, mol%xyz)
       ! ------------------------------------------------------------------------
       !> CONSTRAINTS & SCANS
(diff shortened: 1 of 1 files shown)


marcelmbn:
General question, @Albkat: We don't have any validation tests that access the CLI of `xtb`, i.e. tests that actually execute the `xtb` binary with something like:
```
xtb --hess struc.xyz
```
In this way, we could also test this feature. Up to now, this can easily be broken again by some PR that works on `main.F90` and doesn't know about such features or checks them manually.

marcelmbn:
> General question, @Albkat @awvwgk: We don't have any validation tests that access the CLI of `xtb`, i.e. tests that actually execute the `xtb` binary with a call like the following, right?
> 
> ```
> xtb --hess struc.xyz
> ```
> 
> In this way, we could also test this feature. Up to now, this can easily be broken again by some PR that works on `main.F90` and doesn't know about such features and doesn't check them manually.

marcelmbn:
@demonic-daisy @MtoLStoN Maybe you can look quickly into this to check if the sandwich potential thing could be affected by this (but I don't think so).

marvinfriede (approved):

PR #1177, opened 2025-02-05
Update `model_hessian.f90`

foxtran:
This PR updates model_hessian.f90 in the following way:
1. defines `kl` variable. May affect how torsions Hessian is constructed
2. removes all dcopy since they are used for 3-elements copy. Can be done at Fortran level
3. Partially removes go-to statements


toxtran:
@marvinfriede, can we merge it?

marvinfriede (approved):

PR #1179, opened 2025-02-05
Enable hyphen separated range of atoms to modify mass

marcelmbn:
### Situation before 
before, only single entries like:
```
$hess
modify mass: 12, 2.0, 3, 4.0
scale mass: 2, 1.5, 5, 6.0
$end
```
were possible.

### Improvement
Hyphen-separated ranges of atoms can be read in, like:
```
$hess
scale mass: 12-50,2.0
modify mass: 1-3,400.0
$end
```

### DIFF
--- a/src/constrain_param.f90
+++ b/src/constrain_param.f90
@@ -1242,7 +1242,7 @@ subroutine set_hess(env,key,val,nat,at,idMap,xyz)
    type(TAtomList) :: atl
    integer, allocatable :: list(:)
    real(wp) :: ddum
-   integer  :: i,j,idum,ati,narg
+   integer  :: i,j,idum,iat,narg
    character(len=p_str_length),dimension(p_arg_length) :: argv
    character(len=256) :: warningstring
 
@@ -1286,16 +1286,16 @@ subroutine set_hess(env,key,val,nat,at,idMap,xyz)
             endif
             call atl%to_list(list)
             do idum = 1, size(list)
-               ati = list(idum)
-               if (ati.gt.nat) then
+               iat = list(idum)
+               if (iat.gt.nat) then
                   write(warningstring, '(a, i0, a)') 'Attempted setting atom mass for atom ', &
-                     & ati, ' that is not present in system.'
+                     & iat, ' that is not present in system.'
                   call env%warning(trim(warningstring), source)
                   cycle
                endif
-               atmass(ati) = ddum
+               atmass(iat) = ddum
                write(env%unit,'(a,1x,i0,1x,a,1x,g0)') &
-                  & 'mass of atom ',ati,' changed to',atmass(ati)
+                  & 'mass of atom ',iat,' changed to',atmass(iat)
             enddo
             call atl%destroy()
          endif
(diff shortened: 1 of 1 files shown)


src/constrain_param.f90
```diff
@@ -1238,14 +1239,12 @@ subroutine set_hess(env,key,val,nat,at,idMap,xyz)
...
-   integer  :: idum
-   real(wp) :: ddum
-   logical  :: ldum
-   integer  :: i,j
+   type(TAtomList) :: atl
    integer, allocatable :: list(:)
-
-   integer  :: narg
+   real(wp) :: ddum
+   integer  :: i,j,idum,ati,narg,len_warning
```
marvinfriede, line 1245: `len_warning` is  not used.
```suggestion
   integer  :: i,j,idum,ati,narg
```

src/constrain_param.f90
```diff
@@ -1279,15 +1278,26 @@ subroutine set_hess(env,key,val,nat,at,idMap,xyz)
...
-         if (getValue(env,trim(argv(i)),idum).and.&
-             getValue(env,trim(argv(j)),ddum)) then
-            if (idum.gt.nat) then
-               call env%warning('Attempted setting atom mass not present in system.',source)
-               cycle
+         if (getValue(env,trim(argv(j)),ddum)) then
+            call atl%new(argv(i))
+            if (atl%get_error()) then
+               call env%warning('something is wrong in the mass list',source)
+               return
```
marvinfriede, line 1285: Before, we just cycled. Is the return here intended?

marvinfriede (approved):

src/constrain_param.f90
marcelmbn: We still cycle (that was shifted to another part of the code block some lines below) in case `ati` is higher than the number of atoms in the system. 
The referenced check throws an error if the syntax of the atom list is unexpected and in this case, we don't read the rest. But we could also `cycle` at this point and try to read the other entries of the comma-separated list.

src/constrain_param.f90
```diff
@@ -1279,15 +1278,26 @@ subroutine set_hess(env,key,val,nat,at,idMap,xyz)
...
-            write(env%unit,'(a,1x,i0,1x,a,1x,g0)') &
-               'mass of atom ',idum,' changed to',atmass(idum)
+            call atl%to_list(list)
+            do idum = 1, size(list)
+               ati = list(idum)
+               if (ati.gt.nat) then
+                  write(warningstring, '(a, i0, a)') 'Attempted setting atom mass for atom ', &
+                     & ati, ' that is not present in system.'
+                  call env%warning(trim(warningstring), source)
+                  cycle
```
marcelmbn, line 1294: This is the part from above; now happens here. 🙂

marcelmbn:
Renamed `ati` to `iat` to prevent confusion with atom type. (Variable represents the numbered atom in the range from `iat = 1,nat`, and not the type of atom, i.e. the proton number)

src/constrain_param.f90
marcelmbn: Now `cycle` instead of `return`.

marvinfriede: Cycle makes more sense if we only emit a warning imo.

marvinfriede (dismissed):

marvinfriede (approved):

PR #1183, opened 2025-02-05
Define types of all variables 

foxtran:
This PR defines all unknown types to integer where it was necessary. Now, code can be compiled with `gfortran -fimplicit-none`.

### DIFF
--- a/src/prmat.f90
+++ b/src/prmat.f90
@@ -95,6 +95,7 @@
       end subroutine preig2
 
       SUBROUTINE PREIG3(IO,E,NORBS)
+      IMPLICIT INTEGER (I-N)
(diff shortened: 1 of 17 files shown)


src/prmat.f90
```diff
@@ -300,8 +303,10 @@ SUBROUTINE PRMAT4(IUOUT,R,N,M,HEAD)
       END
 
       SUBROUTINE PRMATI(IUOUT,RR,N,M,HEAD)
+      IMPLICIT INTEGER(I-N)
       CHARACTER*(*) HEAD
       integer RR(*)
+      REAL*4 R
```
foxtran, line 309: Assumed from used float intrinsic.

awvwgk (approved):

PR #1187, opened 2025-02-07
Clarified the aISS output

cplett:
The `fast mode` printed in the output does not refer to the `--fast` flag, but to the use of the ATM term. This was clarified with the commit.

### DIFF
--- a/src/docking/search_nci.f90
+++ b/src/docking/search_nci.f90
@@ -199,7 +199,7 @@ contains
       write (env%unit, *) '|         Starting Energy Screening          |'
       write (env%unit, *) '=============================================='
       write (env%unit, *)
-      if (.not. fulle) write (env%unit, *) ' Fast Mode selected (recommended)'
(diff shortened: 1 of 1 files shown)


src/docking/search_nci.f90
```diff
@@ -199,7 +199,7 @@ END SUBROUTINE Quicksort
       write (env%unit, *) '|         Starting Energy Screening          |'
       write (env%unit, *) '=============================================='
       write (env%unit, *)
-      if (.not. fulle) write (env%unit, *) ' Fast Mode selected (recommended)'
+      if (.not. fulle) write (env%unit, *) ' No ATM term employed (recommended)'
       if (.not. fulle) write (env%unit, *) ' If ATM term should be included, use -atm option.'
```
marvinfriede, line 203: Is it `-atm` or `--atm`? Or does both work? The docs say two dashes.

cplett: Both works as for most of the `xtb` flags.

marvinfriede (approved):

PR #1191, opened 2025-02-11
Update OpenMP parallelization in WSC to avoid data-races

foxtran:
This PR removed data-races from shared minpos/mindist/nminloc/nmindist. Now, `wsc` and `peeq` tests should be repeatable.

### DIFF
--- a/src/generate_wsc.f90
+++ b/src/generate_wsc.f90
@@ -54,12 +54,6 @@ subroutine generate_wsc(mol,wsc)
    call wsc%allocate(mol%n,rep)
 
 ! ------------------------------------------------------------------------
-! initialize
-! ------------------------------------------------------------------------
-   allocate( lattr(3,wsc%cells),      source = 0 )
-   allocate( dist(wsc%cells),         source = 0.0_wp )
-   allocate( trans(wsc%cells),        source = .true. )
-! ------------------------------------------------------------------------
 ! Create the Wigner-Seitz Cell (WSC)
 ! ------------------------------------------------------------------------
    wsc%at  = 0
@@ -69,6 +63,10 @@ subroutine generate_wsc(mol,wsc)
    !$omp private(ii,jj,wc,c,dist,trans,t,lattr,rw) &
    !$omp shared(mol,wsc,rep) &
    !$omp private(mindist,minpos,nmindist,nminpos)
+
+   allocate( lattr(3,wsc%cells),      source = 0 )
+   allocate( dist(wsc%cells),         source = 0.0_wp )
+   allocate( trans(wsc%cells),        source = .true. )
 
    ! Each WSC of one atom consists of n atoms
    !$omp do collapse(2) schedule(dynamic,32)
--- a/src/generate_wsc.f90
+++ b/src/generate_wsc.f90
@@ -127,6 +127,7 @@ subroutine generate_wsc(mol,wsc)
       end do
    end do
    !$omp end do
+   deallocate(lattr, dist, trans)
    !$omp end parallel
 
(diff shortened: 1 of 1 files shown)


foxtran:
Just made a small update to make allocation in parallel region for avoiding OMP stack usage.

foxtran:
It fixes random failures of `xtb/gfnff` test also :-)

src/generate_wsc.f90
```diff
@@ -54,24 +54,27 @@ subroutine generate_wsc(mol,wsc)
...
+   allocate( dist(wsc%cells),         source = 0.0_wp )
+   allocate( trans(wsc%cells),        source = .true. )
+
    ! Each WSC of one atom consists of n atoms
+   !$omp do collapse(2) schedule(dynamic,32)
    do ii=1,mol%n
       do jj=1,mol%n
+         lattr = 0
+         dist = 0.0_wp
+         trans = .true.
```
marvinfriede, line 77: Why do we need to set the arrays again here? As far as I can see, they are private and properly initialized before.

src/generate_wsc.f90
```diff
@@ -54,24 +54,27 @@ subroutine generate_wsc(mol,wsc)
...
          !if (ii.eq.jj) cycle
          ! find according neighbours
          c=0
```
marvinfriede, line 80: Same here

src/generate_wsc.f90
```diff
@@ -54,24 +54,27 @@ subroutine generate_wsc(mol,wsc)
...
    wsc%at  = 0
    wsc%itbl= 0
-!$omp parallel default(none) &
-!$omp private(ii,jj,wc,c,dist,trans,t,lattr,rw) &
-!$omp shared(mol,wsc,rep) &
-!$omp shared(mindist,minpos,nmindist,nminpos)
-!$omp do schedule(dynamic)
+
+   !$omp parallel default(none) &
+   !$omp private(ii,jj,wc,c,dist,trans,t,lattr,rw) &
```
marvinfriede, line 63:
```suggestion
   !$omp private(ii,jj,wc,aa,bb,cc,c,dist,trans,t,lattr,rw) &
```

marvinfriede:
Can you also remove the unused `ich` variable?

src/generate_wsc.f90
```diff
@@ -123,7 +126,7 @@ subroutine generate_wsc(mol,wsc)
          endif
       end do
    end do
-!$omp enddo
-!$omp endparallel
+   !$omp end do
+   !$omp end parallel
```
marvinfriede, line 130: While not strictly necessary, I think it does not hurt to be extra clear about the deallocation in the parallel region (as in [`tblite`](https://github.com/tblite/tblite/blob/ba9c91e8708b332fad67bcb021c7054b544ef7fb/src/tblite/coulomb/charge/effective.f90#L560-L562)).

```suggestion
   !$omp end do
   deallocate(dist, lattr, trans)
   !$omp end parallel
```

src/generate_wsc.f90
foxtran: In allocate statement, they will be initialized once when thread will start to work.

src/generate_wsc.f90
foxtran: Got it. Removed.

foxtran:
@marvinfriede, I have updated the code with your suggestions

src/generate_wsc.f90
marvinfriede: Ah, I see these variables need to be reset.

marvinfriede (approved):

PR #1205, opened 2025-02-27
Increase threshold for lowest frequencies

### LINKED ISSUE #1203: Segmentation fault for scan / constrained optimization for diatomic molecules
xtb version: pre-compiled binary xtb-6.7.1-linux-x86_64 

**Describe the bug**
When doing a scan / constrained optimization for diatomic molecules like H2 O2, the xtb program will execute till showing the ANCOPT SETUP and 
```
generating ANC from exact Hessian ...
Using Lindh-Hessian (1995)

```
and then give the following error message:
```
xtb                000000000046E88A  Unknown               Unknown  Unknown
xtb                0000000002D1BD40  Unknown               Unknown  Unknown
xtb                000000000109EF48  xtb_detrotra_mp_d         141  detrotra.f90
xtb                0000000000FBB267  xtb_type_anc_mp_g         244  anc.f90
xtb                0000000000B14B06  xtb_optimizer_mp_         532  optimizer.f90
xtb                0000000000A2754E  xtb_geoopt_mp_geo         147  geoopt_driver.f90
xtb                000000000042AE6C  xtb_prog_main_mp_         832  main.F90
xtb                0000000000450EA2  MAIN__                     57  primary.f90
xtb                0000000000401E42  Unknown               Unknown  Unknown
xtb                0000000002D1298A  Unknown               Unknown  Unknown
xtb                0000000002D14227  Unknown               Unknown  Unknown
xtb                0000000000401D05  Unknown               Unknown  Unknown
```

**To Reproduce**
```
xtb --gfn 1 --opt extreme --input scan.inp --verbose h2_opt.xyz > scan.txt
```
scan.inp:
```
$constrain
 force constant=10.0
 distance: 1, 2, 0.7
$scan
 1: 0.7, 1.0, 31
$end
```
h2_opt.xyz:
```
2
 energy: -1.036708103979 gnorm: 0.000000077126 xtb: 6.7.1 (edcfbbe)
H            0.01720557368941       -0.00000009008817       -0.00000006005878
H            0.78279442629475       -0.00000409870203       -0.00000273246802
```

[scan.txt](https://github.com/user-attachments/files/18992790/scan.txt)

foxtran:
Fixes #1203 

H2 eigvals are the following:
```
2.9933945092181844E-015   8.5529038621230392E-015   2.6862916288779701E-014   3.4268547221618851        3.4268547221618966        40.610372376753845
```

### DIFF
--- a/src/detrotra.f90
+++ b/src/detrotra.f90
@@ -43,7 +43,7 @@ module xtb_detrotra
   
     nn = 0
     do ii=1, n3
-       if(eig(ii).gt.0.05) cycle  ! only lowest checked
+       if(eig(ii).gt.5.0_wp) cycle  ! only lowest checked
   
        kk=0
        do j=1,mol%n


foxtran:
@marvinfriede, ping :-)

thfroitzheim:
Isn't this the same issue we see in #1210. For linear water, it appears to be around 900 wavenumbers if I see it correctly. Why do we limit this at all? Shouldn't the ranked value be correct even if we check for all modes? If not, we need to introduce a check that at least 5 or 6 values are included for linear and non-linear molecules. Otherwise, it will fail again, right?

foxtran:
> Why do we limit this at all?

Now, there is no limits. All freqs will be checked.

> Isn't this the same issue we see in https://github.com/grimme-lab/xtb/pull/1210.

They are different.

foxtran:
Ah.. It always checks 5/6 freq., but only 3 can be imaginary :(

thfroitzheim:
Are we actually sure that this is the right way of finding rotations and translations? As far as I know, one normally projects rotations and translations out so that the final hessian has 5/6 degrees of freedom less. The remaining degrees of freedom are then all vibrations. We have code to do something like this [here](https://github.com/grimme-lab/xtb/blob/main/src/freq/project.f90), but we still try to assign frequencies to the full hessian. Maybe it is better to do it properly via a projection to make sure it works in all cases.

foxtran:
It seems to me that this part was never adjusted to such cases. So, let it be as is without significant updating like in #1210

PR #1206, opened 2025-02-28
Introduce ifx

### LINKED ISSUE #908: ifx integration
The [deprecation](https://community.intel.com/t5/Blogs/Tech-Innovation/Tools/Deprecation-of-The-Intel-Fortran-Compiler-Classic-ifort/post/1541699)  Intel Fortran Compiler Classic (`ifort`)

[Porting](https://www.intel.com/content/www/us/en/developer/articles/guide/porting-guide-for-ifort-to-ifx.html) to the LLVM-based Intel Fortran Compiler `ifx` 

**Additional context**
Since the direct build with the swapping of compiler name (`ifort` -> `ifx` | `icc` -> `icx`) doesn't seem to succeed, one has to rewrite the corresponding build scripts and establish a CI routine.

foxtran:
Closes #908 

It is not clear how does it work outside of test suite.

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -243,6 +243,55 @@ jobs:
            with:
             name: xtb-bleed.tar.xz
             path: xtb-bleed.tar.xz
+   intel-ifx-meson-build:
+      runs-on: ${{ matrix.os }}
+      strategy:
+         fail-fast: false
+         matrix:
+            os: [ubuntu-latest]
+            toolchain:
+              - {compiler: intel, version: '2024.1'}
+              - {compiler: intel, version: '2025.0'}
+      env:
+         APT_PACKAGES: >-
+            intel-oneapi-mkl-2022.1.0 intel-oneapi-mkl-devel-2022.1.0
+      steps:
+         - name: Checkout code
+           uses: actions/checkout@v4
+         - name: Setup Fortran
+           uses: fortran-lang/setup-fortran@v1
+           with:
+            compiler: ${{ matrix.toolchain.compiler }}
+            version: ${{ matrix.toolchain.version }}
+         - name: Setup Python
+           uses: actions/setup-python@v5
+           with:
+            python-version: 3.x
+         - run: pip3 install meson ninja --user
+         - name: Add Intel repository
+           if: contains(matrix.os, 'ubuntu')
+           run: |
+            wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+            sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
+            rm GPG-PUB-KEY-INTEL-SW-PRODUCTS-2023.PUB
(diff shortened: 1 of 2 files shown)


thfroitzheim:
There appears to be a problem with MKL if you look at the meson setup: 

```
xtb 6.7.1

  Subprojects
    cpx               : YES
    dftd4             : NO
                        Problem encountered: MKL not supported for this compiler
    json-fortran-8.2.5: YES (from mctc-lib)
    mctc-lib          : YES
    mstore            : YES (from cpx => numsa)
    multicharge       : NO
                        Problem encountered: MKL not supported for this compiler
    numsa             : YES (from cpx)
    tblite            : NO
                        Problem encountered: MKL not supported for this compiler
    test-drive        : YES (from cpx => toml-f)
    toml-f            : YES (from cpx)

  User defined options
    default_library   : static
    fortran_link_args : -lifcoremt -static
    lapack            : mkl
    libdir            : lib
    prefix            : /
```
    
This way `dftd4`, `multicharge`, and `tblite` cannot be compiled with ifx. 

I think this looks similar to the problem seen already in `dftd4` [here](https://github.com/dftd4/dftd4/issues/252) and [here](https://github.com/dftd4/dftd4/issues/253). It is not in version 3.6.0 of `dftd4` but was fixed in 3.7.0. If this fixes the problem, we should adapt the solution also to multicharge and tblite.

foxtran:
@thfroitzheim, not surprised that it fails:

https://github.com/tblite/tblite/blob/5000ff61a76b26b48223bf6ce7e54d00699ab833/config/meson.build#L59-L71

tblite does not have a support of IFX compiler. More specifically it does not support compilers other that ifort/gcc. I do not think that it is a xtb problem.

thfroitzheim:
I completely agree. But to really support ifx for xtb, we need to expand the ifx support also to the subprojects (especially tblite, which is actually used). 

Still, this is a great start!

.github/workflows/fortran-build.yml
```diff
@@ -243,6 +243,47 @@ jobs:
            with:
             name: xtb-bleed.tar.xz
             path: xtb-bleed.tar.xz
+   intel-ifx-meson-build:
+      runs-on: ${{ matrix.os }}
+      strategy:
+         fail-fast: false
+         matrix:
+            os: [ubuntu-latest]
```
thfroitzheim, line 251: Can we do this for both meson and cmake or is there a problem with ifx and cmake?

foxtran: No, there is no issues: I'm using CMake mostly :-)
I'm just lazy to implement the same stuff for CMake + IFX :)

thfroitzheim: Ok given that we also test ifort only for meson this should be fine.

thfroitzheim:
Since it works fine, we can merge this as a first step to fully support ifx and do the subprojects later.

thfroitzheim (dismissed):

.github/workflows/fortran-build.yml
foxtran: I have added IFX + CMake :-)

thfroitzheim (approved):
LGTM. Even better with the cmake test.

Issue #986, opened 2024-03-07
Bug in gradient calculation


hanimustafa:
**Describe the bug**
There is a bug in the gradient calculation in XTB. Enabling the debug numerical differentiation code confirms it.  Bizarrely, the error seems to be in the Z component of only 2 of the carbons (!!). This bug has been sitting with me for a while, but after procrastinating the bug hunting for too long I decided to file it instead.

**To Reproduce**
Run XTB against the structure provided. The gradient norm should be around 0.0059248, instead XTB calculates it as 0.0065176. Enable the debug numerical differentiation in order to confirm it. I attach the output with the debug code enabled.

[output.log](https://github.com/grimme-lab/xtb/files/14529324/output.log)
[NEW2-vstrict-valinomycin-xtb.xyz.log](https://github.com/grimme-lab/xtb/files/14529335/NEW2-vstrict-valinomycin-xtb.xyz.log)

foxtran:
There is a much smaller structure with the same bug
```
13
symmetry c1
C       -0.495237121      0.340532816      3.198195224
C        0.114762879     -2.074467184      1.869195224
O        0.709762879      0.098532816      2.696195224
C        1.151762879     -1.279467184      2.696195224
O       -1.178237121     -0.490467184      3.742195224
O       -0.422237121     -1.512467184      0.937195224
C        1.527762879     -1.673467184      4.120195224
H        2.058762879     -1.268467184      2.077195224
H        2.259762879     -2.478467184      4.107195224
H        1.991762879     -0.812467184      4.602195224
H        0.658762879     -1.963467184      4.707195224
H       -0.887516562      1.343243926      3.114009292
H       -0.124377242     -3.098899245      2.113635305
```

foxtran:
And then for...
```
2
symmetry c1
O        0.709762879      0.098532816      2.696195224
C        1.151762879     -1.279467184      2.696195224
```

Printout:
```
analytical gradient
    1    -0.087614239593840     0.273150276380796     0.000000000000000
    2     0.087614239593840    -0.273150276380796    -0.000000000000000

numerical gradient
    1    -0.087614242083589     0.273150279284096     0.000000000000000
    2     0.087614238530875    -0.273150274843204    -0.000000004884981

difference gradient
    1     0.000000002489749    -0.000000002903299     0.000000000000000
    2     0.000000001062965    -0.000000001537593     0.000000004884981
```

foxtran:
Even water molecule is affected :|

[...]

PR #1214, opened 2025-03-11
Introduce H0 without overlap to remove division by S

thfroitzheim:
Fixes the remaining issues with the gradient in #986. I added an additional array for H0 without the overlap to avoid the division by the overlap later for the gradient (this loses information of either H0 or S is zero). This is not a nice solution but an intermediate fix before replacing everything with tblite. 

The gradient is now correct w.r.t. both the numerical gradient and tblite. As a result, the hessian changes even in the tests.


foxtran:
I wonder if H0 really need to keep information about overlap.

thfroitzheim:
I thought so too but the problem is the remainder of the periodic implementation. During the construction of H0, we iterate over the images in the adjoining unit cells, which we cannot do later in the SCF. This is why tblite calculates H0 again in the Hamiltonian gradient. 

We could rework the whole construction here, but I don't see the point, given that there is a correct implementation in tblite.

awvwgk:
When implementing tblite I found that using H0 to reconstruct the Hamiltonian integral in the gradient calculation was limiting especially for periodic boundary conditions. In xtb it should not be hard to introduce the same approach but a lot of extra work if you plan to change this part soon anyway.

Albkat (approved):
Lgtm

Issue #1120, opened 2024-10-26
xtb's dependency hell


foxtran:
xtb has multiple dependencies and those also have their own dependensies... Their versions are managed via `subprojects/*.wrap` files which are not unified between different deps sometimes.

**To Reproduce**
1. clone xtb
2. configure for example with cmake via command `cmake -Bbuild -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DCMAKE_Fortran_COMPILER=gfortran`
3. run
```
   head -n 10 `find . -name "*.wrap" | rev | sort | rev | grep subprojects`
```

One will see something like:
```
==> ./subprojects/toml-f.wrap <==
[wrap-git]
directory = toml-f
url = https://github.com/toml-f/toml-f
revision = v0.4.1

==> ./build/_deps/s-dftd3-src/subprojects/toml-f.wrap <==
[wrap-git]
directory = toml-f
url = https://github.com/toml-f/toml-f
revision = v0.2.4

==> ./build/_deps/tblite-src/subprojects/toml-f.wrap <==
[wrap-git]
directory = toml-f
url = https://github.com/toml-f/toml-f
revision = v0.4.0
```

So, s-dftd3, tblite and xtb have different versions of toml-f. At build/_deps/toml-f-src, I have v0.4.0, not v0.4.1 or v0.2.4... So, it is not clear which version will be downloaded first :) I suppose for toml-f it is not so critical, but for dftd4 it is because old v3.4.0 was selected.

**Expected behaviour**
All deps must have the same version

marvinfriede:
We should also consider pinning tblite to a release version and not to HEAD (both meson and cmake).

thfroitzheim:
We have to add the revision, which we pass for each dependency, also to the utility function similar to [tblite](https://github.com/tblite/tblite/blob/main/config/cmake/tblite-utils.cmake).

Issue #1091, opened 2024-08-22
compilation of `xtb` 6.7.1 failing


boegel:
**Describe the bug**

compilation of `xtb` 6.7.1 is failing with GCC 12.3.0 is failing with:

```
[526/899] Compiling Fortran object libxtb.so.6.7.1.p/src_dipro_xtb.F90.o
FAILED: libxtb.so.6.7.1.p/src_dipro_xtb.F90.o libxtb.so.6.7.1.p/xtb_dipro_xtb.mod
gfortran -Ilibxtb.so.6.7.1.p -I. -I../xtb-6.7.1 -I../xtb-6.7.1/include -Isubprojects/mctc-lib/libmctc-lib.a.p -Isubprojects/json-fortran-8.2.5/libjsonfortran.a.p -I../xtb-6.7.1/subprojects/tblite/include -Isubprojects/tblite/libtblite.a.p -Isubprojects/dftd4/libdftd4.a.p -I../xtb-6.7.1/subprojects/dftd4/include -Isubprojects/multicharge/libmulticharge.a.p -Isubprojects/s-dftd3/libs-dftd3.a.p -I../xtb-6.7.1/subprojects/s-dftd3/include -Isubprojects/toml-f/libtoml-f.a.p -Isubprojects/numsa/libnumsa.a.p -Isubprojects/cpx/libcpx.a.p -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -O2 -fdefault-real-8 -fdefault-double-8 -ffree-line-length-none -fbacktrace -O2 -ftree-vectorize -march=native -fno-math-errno -fPIC -fopenmp -pthread -fopenmp -DWITH_TBLITE=1 -DWITH_CPCMX=1 -Jlibxtb.so.6.7.1.p -o libxtb.so.6.7.1.p/src_dipro_xtb.F90.o -c ../xtb-6.7.1/src/dipro/xtb.F90
../xtb-6.7.1/src/dipro/xtb.F90:53:42:

   53 |       call new_gfn2_calculator(xcalc, mol)
      |                                          1
Error: Missing actual argument for argument error at (1)
../xtb-6.7.1/src/dipro/xtb.F90:55:42:

   55 |       call new_gfn1_calculator(xcalc, mol)
      |                                          1
Error: Missing actual argument for argument error at (1)
../xtb-6.7.1/src/dipro/xtb.F90:57:43:

   57 |       call new_ipea1_calculator(xcalc, mol)
[...]

Issue #981, opened 2024-03-05
test suite of a source build fails for 6.7.0 


sheepforce:
**Describe the bug**
xTB 6.7.0 fails 4 test cases (`10 - xtb/gfn1`, `11 - xtb-gfn2`, `13 - xtb/hessian` and `17 - xtb/oniom`) with large numerical deviations.

Building with:

  * tblite, commit 82c3dcb9f76f1b45639b7a17c27e7ba40fffe472
  * CPCM-X, commit 7de0d7be85a10a19d60220a8d25eaa750f282019
  * simple-dftd3-1.0.0
  * multicharge-0.2.0
  * openblas-0.3.26 (wrapped in standard interface)
  * mctc-lib-0.3.1
  * toml-f-0.4.2
  * test-drive-0.4.0
  * cmake-3.28.2
  * gfortran-13.2.0

**To Reproduce**
Steps to reproduce the behaviour:

1. ~~happens with input (include input files)~~ ctest test suite fails. Full build log is here: [xtb-build.txt](https://github.com/grimme-lab/xtb/files/14493427/xtb-build.txt)
2. ~~start `xtb` with (all the options here)~~
3. ~~run `xtb` with your options and the `--verbose` flag~~
4. ~~output showing the error~~ (see 1)

You can exactly reproduce this build by running `nix build github:nix-qchem/nixos-qchem/xtb#xtb -L`

**Expected behaviour**
The test suite should pass.

**Additional context**
I'm maintaining the xTB + tblite ecosystem for the Nix package manager and require reproducible builds from source for all packages.

sheepforce:
> cmake-3.28.2
> gfortran-13.2.0

CMake generates standard Makefiles without additional arguments, so plain Make :)

Albkat:
I see, I tested this with meson/ninja/ifort combination.
I will look into it:)

Albkat:
All right, I can reproduce this, but it will probably take some time to debug

sheepforce:
I've switched over to a meson+ninja build for all packages and can confirm that the test suite passes for this build type but otherwise the same dependencies.

toxtran:
This issue should be fixed via #1121.

@sheepforce, could you please reckeck?

sheepforce:
> This issue should be fixed via #1121.
> 
> @sheepforce, could you please recheck?

[...]

PR #1215, opened 2025-03-11
Fix the dependencies on their newest version for meson and cmake

thfroitzheim:
Fixes the dependency versions and should avoid issues such as #1091 and #981. The dependencies should now also be consistent for all subprojects (see #1120) 

Only numsa and CPCM-X remain on `head` since there is either no released version or the release is too old.


awvwgk:
> Only numsa and CPCM-X remain on head since there is either no released version or the release is too old.

How about tagging a new release for those projects then?

thfroitzheim:
I released both numsa and CPCM-X. Now all the dependencies are consistent (this should close #1120). During this update, I removed all unnecessary wrap files and placed the automatically generated wrap files in the gitignore. I don't really see a reason to keep them as we are not using these dependencies but maybe I am missing something @awvwgk.

marvinfriede (approved):

PR #1217, opened 2025-03-12
Fixed race condition in ceh numgrad

lukaswittmann:
Numerical gradient using any omp parallelization leads to completely different results due to a race condition due to the shared `mol%xyz`. This is now fixed using a private copy.


src/tblite/calculator.F90
```diff
@@ -474,18 +474,18 @@ subroutine num_grad_chrg(env, mol, tblite)
    step2 = 0.5_wp / step
    call get_ceh(env,mol,tblite)
    
-   !$omp parallel do private(j,cehr,cehl) shared(env, numgrad, mol, tblite, step, step2)
```
foxtran, line 477: It would be nice to add `default(none)`

foxtran (changes requested):

src/tblite/calculator.F90
```diff
@@ -474,18 +474,18 @@ subroutine num_grad_chrg(env, mol, tblite)
    step2 = 0.5_wp / step
    call get_ceh(env,mol,tblite)
    
-   !$omp parallel do private(j,cehr,cehl) shared(env, numgrad, mol, tblite, step, step2) 
+   !$omp parallel do private(j, cehr, cehl, local_mol) shared(env, numgrad, mol, tblite, step, step2)
    do i = 1, mol%n
+      local_mol = mol    ! create a thread-private copy of mol
```
foxtran, line 479: Probably, you want to `call local_mol%copy(mol)`. At least it is done in this way in `src/type/calculator.f90`

lukaswittmann:
Thanks @foxtran, implemented both suggestions!

src/tblite/calculator.F90
marvinfriede: In general, this is not necessary, because the Molecule type only contains allocatable arrays and no pointer references. And the underlying data of arrays is copied when you do an assignment. However, the assignment operator is overwritten for the Molecule type. Not sure how this influences this.

src/tblite/calculator.F90
```diff
@@ -464,7 +464,7 @@ subroutine num_grad_chrg(env, mol, tblite)
    real(wp) :: numgrad(3, mol%n, mol%n)
    
    real(wp), allocatable, dimension(:) :: cehr, cehl
-   !
+   type(TMolecule) :: local_mol
    integer :: i,j,k, ich
```
marvinfriede, line 468:
```suggestion
   integer :: i, j
```

foxtran:
@lukaswittmann, try to use `git rebase` for avoiding so strange commit history.

You can try to do in this way:
```
git checkout ceh-numgrad-omp-fix # take you branch
git remote add lab git@github.com:grimme-lab/xtb.git # add another remote repo, here it is a main repo
git fetch lab # get all data from main repo
git rebase lab/main # rebase on top of main branch from main repo
```
I hope I did not make a typo :)

marvinfriede (dismissed):

src/tblite/calculator.F90
marvinfriede: Oops, I did not see that the subroutine continues. My bad.

src/tblite/calculator.F90
```diff
@@ -464,8 +464,8 @@ subroutine num_grad_chrg(env, mol, tblite)
    real(wp) :: numgrad(3, mol%n, mol%n)
    
    real(wp), allocatable, dimension(:) :: cehr, cehl
-   !
-   integer :: i,j,k, ich
+   type(TMolecule) :: local_mol
+   integer :: i, j
```
marvinfriede, line 468:
```suggestion
   integer :: i, j, k, ich
```

marvinfriede (approved):

src/tblite/calculator.F90
```diff
@@ -474,18 +474,18 @@ subroutine num_grad_chrg(env, mol, tblite)
    step2 = 0.5_wp / step
    call get_ceh(env,mol,tblite)
    
-   !$omp parallel do private(j,cehr,cehl) shared(env, numgrad, mol, tblite, step, step2) 
+   !$omp parallel do private(j, cehr, cehl, local_mol) shared(env, numgrad, mol, tblite, step, step2) default(none)
    do i = 1, mol%n
+      call local_mol%copy(mol) ! create a thread-private copy of mol
       do j = 1, 3
```
foxtran, line 480: Probably, it will not work with IFX

foxtran: @lukaswittmann, could you please provide a test for ceh hessian with TBLITE?

marvinfriede: Why?

lukaswittmann: There is already a test for the numerical gradient in tblite.

foxtran: It has the same pattern like in #1206. By some reason, there were segfault during creating openmp section with IFX.

foxtran: How can I start the problematic code? I'm using command `xtb inp.xyz --ceh --tblite --hess` but it does not look that it goes to that place.

foxtran: It works with IFX :)

foxtran (approved):

src/tblite/calculator.F90
lukaswittmann: You can execute this part via
```
xtb coord.xyz -P <N> --ceh grad 0.001 --acc 0.010000 
```
It calcualted the numerical CEH gradient, not the numerical hessian.

PR #1219, opened 2025-03-13
Solvation printout, fix for #1185

### LINKED ISSUE #1185: Printout: total w/o Gsasa/hb
When using the ALPB or the GBSA solvation models the printout "total w/o Gsasa/hb" is a bit misleading since it also includes Gshift. 

Therefore this printout should be changed to something like: "total w/o Gsasa/hb/Gshift"

This should be changed in the xtb calculator as well as in the gfnff calculator.

lukaswittmann:
This fixed #1185 by now explicitly stating that the second energy is also without Gshift. Due to the formatting the total energy part has to be removed. Does anyone have better ideas how to possibly handle that?

### DIFF
--- a/src/gfnff/calculator.f90
+++ b/src/gfnff/calculator.f90
@@ -284,7 +284,7 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
       if (.not.set%silent.and.allocated(solvation)) then
          write(env%unit,outfmt) "total w/o Gsolv   ", &
             &  results%e_total-results%g_solv, "Eh   "
-         write(env%unit,outfmt) "total w/o Gsasa/hb", &
+         write(env%unit,outfmt) "w/o Gsasa/hb/shift", &
             &  results%e_total-results%g_sasa-results%g_hb-results%g_shift, "Eh   "
       endif
       write(env%unit,outfmt) "gradient norm     ", results%gnorm,  "Eh/a0"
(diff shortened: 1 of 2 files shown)


src/gfnff/calculator.f90
```diff
@@ -284,7 +284,7 @@ subroutine singlepoint(self, env, mol, chk, printlevel, restart, &
       if (.not.set%silent.and.allocated(solvation)) then
          write(env%unit,outfmt) "total w/o Gsolv   ", &
             &  results%e_total-results%g_solv, "Eh   "
-         write(env%unit,outfmt) "total w/o Gsasa/hb", &
+         write(env%unit,outfmt) "w/o Gsasa/hb/shift", &
```
marvinfriede, line 287: Why did you remove the "total"?

lukaswittmann: Due to the formatting (length) of the output. 
Either one would have to change the whole printout (could be problematic for parsing from output) or it would be very ugly.
```
         :::::::::::::::::::::::::::::::::::::::::::::::::::::
         ::                     SUMMARY                     ::
         :::::::::::::::::::::::::::::::::::::::::::::::::::::
         :: total energy              -0.180306225392 Eh    ::
         :: total w/o Gsasa/hb/shift        -0.167096749821 Eh    ::
         :: gradient norm              0.000000000000 Eh/a0 ::
         :: HOMO-LUMO gap              3.214205568594 eV    ::
         ::.................................................::
```
Another option would be to split it into two lines, but I think this solution makes the most sense.

marvinfriede (approved):

PR #1229, opened 2025-03-24
Fix Fukui function for PTB and change spin-state assignment

### LINKED ISSUE #1226: PTB and Fukui
When I attempt to run Fukui index calculations with PTB using the following command, I get an error 

**To Reproduce**
`xtb geom.xyz --ptb --vfukui`

`geom.xyz`
```
3

H 0 0 0
O 0 0 1
H 0 1 1
```

Error:
```
...
Fukui index Calculation

Run single point for reduced species

Run single point for oxidized species

Fukui functions:
     #        f(+)     f(-)     f(0)
     1H       0.000    0.000    0.000
     2O       0.000    0.000    0.000
     3H       0.000    0.000    0.000
########################################################################
[ERROR] Program stopped due to fatal error
-5- Calculation terminated
-4- ptb_calculator_singlepoint: PTB two-step SCF terminated
-3- ptb_calculator_singlepoint: Number of electrons and spin multiplicity do not match.
-2- ptb_calculator_singlepoint: PTB two-step SCF terminated
-1- ptb_calculator_singlepoint: Number of electrons and spin multiplicity do not match.
########################################################################
```

marcelmbn:
### Modified unpaired electron assignment
- The previous spin state assignment set the wavefunction to be open-shell for the anionic case if the original wavefunction has already been open-shell (see the relevant line of code below). If I don't have a major logic issue this seems to have been wrong in general before (`wf_an` is only a copy from the ground-state wavefunction calculated before), and the other way round would be correct (in the simplest case: a neutral closed-shell molecule becomes open-shell in the ionized case). The spin state assignment has been changed to a more reasonable logic.
```Fortran
if (mod(wf_an%wfn%nel,2).ne.0) wf_an%wfn%nopen = 1
```

- Since PTB does not allow for restarting a calculation, assigning the spin state only in the wavefunction does not have any effect. Before, the Fukui calculation crashed since `mol.uhf` didn't match the changed charge anymore (number of electrons was odd in a closed-shell calculation). Here, `mol.uhf` is changed accordingly (in the simplest case; from closed-shell to `mol.uhf=1`).
- Caution: The Fukui results for GFNn-xTB might also change since the `nopen` assignment has been changed.

### Additional changes
- enable PTB parameter loading for Fukui calculations as well
- minor formatting adjustments

Closes #1226.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -550,7 +550,7 @@ contains
                p_run_modef, p_run_mdopt, p_run_metaopt)
             if (set%mode_extrun == p_ext_gfnff) then
                fnv = xfind(p_fname_param_gfnff)
-             elseif (set%mode_extrun .eq. p_ext_ptb) then
+            elseif (set%mode_extrun .eq. p_ext_ptb) then
                   fnv = xfind(p_fname_param_ptb)
             else
                if (set%gfn_method == 0) then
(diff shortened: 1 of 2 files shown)


lukaswittmann (approved):
LGTM!

> The previous spin state assignment set the wavefunction to be open-shell for the anionic case if the original wavefunction has already been open-shell (see the relevant line of code below). If I don't have a major logic issue this seems to have been wrong in general before

I also don't see any major logic issue.

PR #1233, opened 2025-03-25
Do not shift fixed atoms during metaopt

foxtran:
Originally, during randomisation, fixed atoms could be shifted. Now, they will keep their positions.


src/prog/main.F90
```diff
@@ -1156,15 +1157,34 @@ subroutine xtbMain(env, argParser)
             metaset%xyz(:, :, metaset%nstruc) = mol%xyz
             ! randomize structure to avoid zero RMSD
             do i = 1, mol%n
-               do j = 1, 3
-                  call random_number(er)
-                  mol%xyz(j, i) = mol%xyz(j, i) + 1.0e-6_wp * er
-               end do
+               if (.not. fixset%is_fixed(i)) then
+                  do j = 1, 3
+                     mol%xyz(j, i) = mol%xyz(j, i) + normal_distribution(0.1_wp, 0.0_wp)
```
foxtran, line 1162: There is no flag for adjusting displacements :(

src/random_generators.f90
```diff
@@ -0,0 +1,78 @@
...
+! xtb is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
+
+module random_generators
+   use xtb_mctc_accuracy, only: sp, wp
```
awvwgk, line 19: Let's use `sp` and `dp` here.

foxtran: Done

toxtran:
@awvwgk, i'm going to drop this PR since running of omd took too much time for us. Does anyone have interest on it?

foxtran:
Oh... Meson builds are broken :|

PR #1234, opened 2025-03-25
Init some variables in relaxation engine

foxtran:
At 
https://github.com/grimme-lab/xtb/blob/41059b44d84a07dfd99a051e4409ef654faca939/src/relaxation_engine.f90#L1418-L1419
the reading of uninitialized variable may generate SIGFPE. This patch defines `lat_gnorm` as well as other variables related to optimization of cells to avoid such behaviour.


lukaswittmann (approved):
Good catch, looks good to me.

marvinfriede (approved):

PR #1237, opened 2025-03-26
Fix reduced mass calculation

### LINKED ISSUE #367: Possible printing bug for reduced masses from Hessian calculations
Hi!

I am trying to get the reduced masses from a Hessian calculation at the GFN1-xTB level. I get vastly different valuies compared to using the `qcore` implementation of the same method. The frequencies and normal modes are in good agreement, however. Comparing to a reference B3LYP/6-31G(d) calculation I had, the value form the `xtb` implementation also differs substantially, so I think that `qcore` GFN1-xTB is correct, while `xtb` GFN1-xTB might be printing the wrong number.

I am wondering if `xTB` has a printing error, or if there is some kind of hidden conversion factor that I am missing!

The values for the reduced masses for a Hessian calculation for a water molecule are given below: 

`xtb` GFN1-xTB reduced masses [amu]:
`2.1533          1.5401        2.1640`

`qcore` GFN1-xTB reduced masses in [amu]:
`1.0856          1.0425        1.0863`

B6LYP/6-31G(d) values for reference (from an old gaussian calculation) [amu]:
`1.0825                 1.0454                 1.0810`

As you can see, the numbers for `xtb` are very far from the two others. 

All input I used is attached here:

water.xyz

```
3
0 1
O    -0.000337398098   0.372284110820   0.000000000000
H    -0.783368023921  -0.186855235475   0.000000000000
H     0.783705422019  -0.185428880637   0.000000000000
```

Qcore input: `qcore -s "job := hessian(structure(file='water.xyz') xtb())"`

Grimme code input: `xtb water.xyz --gfn 1 --hess`

xtb version 6.3.3 (7355c4f) compiled by 'conda@14e9545d6667' on 2020-09-17

Entos Qcore version: 0.8.10

marcelmbn:
- fix the calculation of the reduced mass to consistent with reference calculations in other implementations as well as DFT reference data (see issue #367)
- more consistent variable names to avoid exploiting variables for different purposes (`isqm` serves also as scratch variable in a later part of the code)
- closes #367
- took Psi4 implementation https://github.com/psi4/psi4/blob/0df59280773a86f7ae1f6cc577731074a0fbaad9/psi4/driver/qcdb/vib.py#L582 as reference and documentation from: https://gaussian.com/vib/

**Normalization** is a relatively straightforward process. Before it is printed out, each of the `3N` elements of `l_{\text{CART},k,i}` is scaled by the normalization factor `N_i` for that particular vibrational mode. The normalization is defined by:

$$
N_i = \sqrt{ \left( \sum_k \left[ l_{\text{CART},k,i} \right]^2 \right)^{-1} }, \quad k = 1 \dots 3N \quad \text{(Eq. 12)}
$$

**Reduced mass** $\mu_i$ for the vibrational mode is calculated similarly ($k = 1 \dots 3N$):

$$
\mu_i = \left( \sum_k \left[ l_{\text{CART},k,i} \right]^2 \right)^{-1}
= \left( \sum_k \left( \frac{l_{\text{MWC},k,i}}{\sqrt{m_k}} \right)^2 \right)^{-1}
= N_i^2 \quad \text{(Eq. 13)}
$$

Note that since $D$ is orthonormal, and we can (and do) choose $L$ to be orthonormal, then $l$ is orthonormal as well.

Since $$D^\top D = I$$ and $$L^\top L = I$$, it follows that $$I = (DL)^\top (DL) = L^\top D^\top D L = L^\top L = I $$

### DIFF
--- a/src/freq/io.f90
+++ b/src/freq/io.f90
@@ -22,7 +22,7 @@ module xtb_freq_io
    private
 
    public :: writeHessianOut, rdhess, wrhess, write_tm_vibspectrum
-   public :: g98fake, g98fake2
+   public :: g98fake, g98fake2, rddipd
 
 contains
 
@@ -78,10 +78,8 @@ contains
       real(wp), intent(out) :: h(nat3, nat3)
       character(len=*), intent(in) :: fname
       integer :: iunit, i, j, mincol, maxcol
-      character(len=5) :: adum
       character(len=80) :: a80
 
-      !     write(*,*) 'Reading Hessian <',trim(fname),'>'
       call open_file(iunit, fname, 'r')
 50    read (iunit, '(a)') a80
       if (index(a80, '$hessian') /= 0) then
@@ -99,6 +97,50 @@ contains
 
 300   return
    end subroutine rdhess
+
+   subroutine rddipd(nat3,dipd,fname)
+   implicit none
+   integer, intent(in)  :: nat3
+   real(wp),intent(out) :: dipd(3,nat3)
+   character(len=*),intent(in) :: fname
+   integer  :: iunit,i,j
+   character(len=80) :: a80
+
+   call open_file(iunit,fname,'r')
+50 read(iunit,'(a)') a80
+   if(index(a80,'$dipgrad').ne.0)then
+      do i=1,nat3
+         read(iunit,*)(dipd(j,i),j=1,3)
+      enddo
+      call close_file(iunit)
+      goto 300
+   endif
+   goto 50
+
+300 return
+   end subroutine rddipd
+
+   subroutine wrdipd(nat3,dipd,fname)
+   use xtb_lin, only : lin
+   implicit none
+   integer, intent(in) :: nat3
+   real(wp),intent(in) :: dipd(3,nat3)
+   character(len=*),intent(in) :: fname
+   integer iunit,i
+   character(len=5)  :: adum
+   character(len=80) :: a80
+
+   adum='   '
+   call open_file(iunit,fname,'w')
+   a80='$dipd'
+   write(iunit,'(a)')a80
+   do i=1,nat3
+      write(iunit,'(a5,3f15.10)')adum,dipd(1:3,i)
+   enddo
+   call close_file(iunit)
+
+   end subroutine wrdipd
+
 
    subroutine write_tm_vibspectrum(ich, n3, freq, ir_int, raman_activity, temp, v_incident)
       use xtb_setparam
--- a/src/freq/io.f90
+++ b/src/freq/io.f90
@@ -49,28 +49,31 @@ contains
    end subroutine writeHessianOut
 
    subroutine wrhess(nat3, h, fname)
-      integer, intent(in) :: nat3
-      real(wp), intent(in) :: h(nat3 * (nat3 + 1) / 2)
-      character(len=*), intent(in) :: fname
-      integer iunit, i, j, mincol, maxcol, k
-      character(len=5) :: adum
-      character(len=80) :: a80
+      !> Arguments
+      integer, intent(in)                 :: nat3
+      real(wp), intent(in)                :: h(nat3 * (nat3 + 1) / 2)
+      character(len=*), intent(in)        :: fname
+      !> Local variables
+      integer                             :: iunit
+      integer                             :: i, j, mincol, maxcol
+      character(len=80)                   :: a80
 
-      adum = '   '
       call open_file(iunit, fname, 'w')
+      ! Write header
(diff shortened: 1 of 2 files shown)


marcelmbn:
I also quickly checked the agreement with a TURBOMOLE reference calculation by taking `hessian` and `dipgrad` from TURBOMOLE, skipping all projection steps and similar, diagonalize `hessian`, bring it together with `dipgrad` and compare the final spectrum and reduced masses with the values resulting from `aoforce` (processed internally with TURBOMOLE):

### `xtb`-processed
```
 projected vibrational frequencies (cm⁻¹)
eigval :       -0.00    -0.00    -0.00     0.00     0.00     0.04
eigval :     1610.11  3868.05  4002.96
 reduced masses (amu)
    1:      1.32    2:      2.57    3:      1.01    4:     16.00    5:      5.64    6:      1.04
    7:      1.09    8:      1.04    9:      1.09
 IR intensities (km·mol⁻¹)
    1:     30.28    2:     11.42    3:     42.26    4:     10.65    5:      0.55    6:     47.56
    7:     95.47    8:     21.03    9:    120.88
```

### TM-processed
```
       mode               7        8        9

     frequency        1610.09  3867.99  4002.90

     symmetry            a        a        a   

        IR               YES      YES      YES
|dDIP/dQ|   (a.u.)     0.0073   0.0034   0.0082
intensity (km/mol)      95.47    21.03   120.88
intensity (  %   )      78.98    17.40   100.00
 
       RAMAN             YES      YES      YES
 
  1   o           x   0.00000  0.00000 -0.07236
                  y   0.00000  0.00000  0.00000
                  z  -0.07186  0.04830 -0.00000
  2   h           x   0.41492  0.59323  0.57432
                  y   0.00000  0.00000  0.00000
                  z   0.57031 -0.38330 -0.40932
  3   h           x  -0.41492 -0.59323  0.57432
                  y   0.00000  0.00000 -0.00000
                  z   0.57031 -0.38330  0.40932

reduced mass(g/mol)     1.085    1.043    1.086
```

src/freq/io.f90
lukaswittmann, line 1: Is there a reason for the go-to statements in `rddipd`?

marcelmbn: Done ✅

src/freq/io.f90
```diff
@@ -49,57 +49,95 @@ subroutine writeHessianOut(fname, hessian)
...
-         call close_file(iunit)
-         goto 300
       end if
-      goto 50
+      call close_file(iunit)
 
-300   return
    end subroutine rdhess
 
+   subroutine rddipd(nat3,dipd,fname)
```
lukaswittmann, line 104: What does this do? A description would be nice.

src/freq/io.f90
```diff
@@ -49,57 +49,95 @@ subroutine writeHessianOut(fname, hessian)
...
+      read(iunit,'(a)') a80
+      if(index(a80,'$dipgrad').ne.0)then
+         do i=1,nat3
+            read(iunit,*)(dipd(j,i),j=1,3)
+         enddo
+      endif
+      call close_file(iunit)
+   end subroutine rddipd
+
+   subroutine wrdipd(nat3,dipd,fname)
```
lukaswittmann, line 122: The same here would be nice.

src/freq/io.f90
```diff
@@ -22,7 +22,7 @@ module xtb_freq_io
    private
 
    public :: writeHessianOut, rdhess, wrhess, write_tm_vibspectrum
-   public :: g98fake, g98fake2
+   public :: g98fake, g98fake2, rddipd, wrdipd
```
lukaswittmann, line 25: Is `wrdipd` needed elsewhere?

lukaswittmann (approved):
Looks good to me, great work!

src/freq/io.f90
marcelmbn: Done ✅

src/freq/io.f90
marcelmbn: Done ✅

src/freq/io.f90
marcelmbn: Three of the four TURBOMOLE format-based routines (`rdhess`, `rddipd`, `wrdipd`) are not "needed" and are mainly relevant for debugging purposes, yet they can be quite helpful. `wrhess` is actually used to dump the `hessian` matrix to a file.

lukaswittmann: I see, thanks for the clarification!

marvinfriede (approved):

Issue #1239, opened 2025-03-27
Improve parallelization of the hydrogen bond potential in GFN-FF calculations


Thomas3R:
Hydrogen bonds (HB) are present in almost every GFN-FF calculation, and the number of HB interactions often becomes very large very quickly (especially for periodic systems). 
The bottleneck are the [gfnff_hbset](https://github.com/grimme-lab/xtb/blob/main/src/gfnff/gfnff_ini2.f90#L764) and [gfnff_hbset0](https://github.com/grimme-lab/xtb/blob/main/src/gfnff/gfnff_ini2.f90#L764) calls, which are potentially called in every energy+gradient calculation of the force field. The HB-list setup in both subroutines is not parallelized and seems to create an overhead (eventually due to bad compiler flags).

Implementing OpenMP parallel loops for the mentioned subroutines should significantly improve the scalability of the force field (for HB-dominated systems).

PR #1240, opened 2025-03-27
Implement parallelization for hydrogen bond list setup in gfnff_hbset

Thomas3R:
As mentioned in #1239, the HB setup benefits from the parallelization implemented here. 
I use local copies of the variables from TNeigh and TGFFNeighbourList types so the constants in the construct (nlist_nhb1, nlist_nhb2, neigh_numctr, neigh_nTrans, hbthr1, hbthr2) can be used as firstprivate.
The `!$omp atomic capture` ensures that each thread works on a different/unique entry (index nlist_nhb1 or nlist_nhb2) of the respective lists.
I confirmed improved scaling on the X23 benchmark and a larger peptide crystal. However, gfnff_hbset0 takes significantly longer and should be parallelized next.

### DIFF
--- a/src/gfnff/gfnff_ini2.f90
+++ b/src/gfnff/gfnff_ini2.f90
@@ -776,75 +776,99 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
       integer i,j,k,nh,ia,ix,lin,ij,inh,jnh, iTri,iTrj,iTrDum
       real(wp) rab,rmsd, rih,rjh
       logical ijnonbond,free
+      ! local copies for omp parallelization
+      integer :: neigh_nTrans, neigh_numctr, nlist_nhb1, nlist_nhb2
 
+      neigh_nTrans = neigh%nTrans
+      neigh_numctr = neigh%numctr
 
       rmsd = sqrt(sum((xyz-nlist%hbrefgeo)**2))/dble(n)
 
       if(rmsd.lt.1.d-6 .or. rmsd.gt. 0.3d0) then ! update list if first call or substantial move occured
       nlist%nhb1=0
       nlist%nhb2=0
+      nlist_nhb1=0
+      nlist_nhb2=0
       ! loop over hb-relevant AB atoms
+      !$omp parallel do default(none) private(i, j, k, nh, ix, iTri, iTrj, iTrDum, rab, &
+      !$omp rih, rjh,ijnonbond,free ) &
+      !$omp firstprivate (nlist_nhb1, nlist_nhb2, neigh_numctr, neigh_nTrans, hbthr1, hbthr2) &
+      !$omp shared(nlist, topo, neigh, xyz)
       do ix=1,topo%nathbAB  
-        i=topo%hbatABl(1,ix) 
-        j=topo%hbatABl(2,ix)
-        do iTri=1, neigh%nTrans ! go through i shifts
-          do iTrj=1, neigh%nTrans ! go through j shifts
-            ! get adjustet iTr -> for use of neigh% distances and bpair with two shifted atoms
-            iTrDum=neigh%fTrSum(neigh%iTrNeg(iTri),iTrj) 
-            if(iTrDum.gt.neigh%nTrans.or.iTrDum.lt.-1.or.iTrDum.eq.0) cycle ! cycle nonsense 
-            rab=NORM2((xyz(:,i)+neigh%transVec(:,iTri))-(xyz(:,j)+neigh%transVec(:,iTrj)))**2
-            if(rab.gt.hbthr1) cycle
-            ! check if ij bonded
-            if(iTrDum.le.neigh%numctr.and.iTrDum.gt.0) then
-              ijnonbond=neigh%bpair(j,i,iTrDum).ne.1
-            else
-              ! i and j are not in neighboring cells
-              ijnonbond=.true. 
-            endif
-            ! loop over relevant H atoms
-            do k=1,topo%nathbH
-              free=.true. ! tripplet not assigned yet
-              nh  =topo%hbatHl(1,k) ! nh always in central cell
-              ! distances for non-cov bonded case
-              rih=NORM2(xyz(:,nh)-(xyz(:,i)+neigh%transVec(:,iTri)))**2
-              rjh=NORM2(xyz(:,nh)-(xyz(:,j)+neigh%transVec(:,iTrj)))**2
-              ! check if i is the bonded A    
-              if(iTri.le.neigh%numctr) then ! nh is not shifted so bpair works without adjustment 
-                if(neigh%bpair(i,nh,iTri).eq.1.and.ijnonbond) then
-                  nlist%nhb2=nlist%nhb2+1
-                  nlist%hblist2(1,nlist%nhb2)=i
-                  nlist%hblist2(2,nlist%nhb2)=j
-                  nlist%hblist2(3,nlist%nhb2)=nh
-                  nlist%hblist2(4,nlist%nhb2)=iTri
-                  nlist%hblist2(5,nlist%nhb2)=iTrj
-                  free=.false. ! not available for nhb1 !!!
-                endif 
-              endif  
-              ! check if j is the bonded A
-              if(iTrj.le.neigh%numctr.and.free) then
-                if(neigh%bpair(j,nh,iTrj).eq.1.and.ijnonbond) then
-                  nlist%nhb2=nlist%nhb2+1
-                  nlist%hblist2(1,nlist%nhb2)=j
-                  nlist%hblist2(2,nlist%nhb2)=i
-                  nlist%hblist2(3,nlist%nhb2)=nh
-                  nlist%hblist2(4,nlist%nhb2)=iTrj
-                  nlist%hblist2(5,nlist%nhb2)=iTri
-                  free=.false. ! not available for nhb1 !!!
-                endif
-              endif  
-              ! check for non-cov bonded A  
-              if(rab+rih+rjh.lt.hbthr2.and.free) then ! sum of rAB,rAH,rBH is below threshold
-                nlist%nhb1=nlist%nhb1+1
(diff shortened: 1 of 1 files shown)


src/gfnff/gfnff_ini2.f90
```diff
@@ -776,83 +776,107 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
+                        ! i and j are not in neighboring cells
+                        ijnonbond=.true.
+                     endif
+                     ! loop over relevant H atoms
+                     do k=1,topo%nathbH
+                        free=.true. ! tripplet not assigned yet
+                        nh  =topo%hbatHl(1,k) ! nh always in central cell
+                        ! distances for non-cov bonded case
+                        rih=NORM2(xyz(:,nh)-(xyz(:,i)+neigh%transVec(:,iTri)))**2
+                        rjh=NORM2(xyz(:,nh)-(xyz(:,j)+neigh%transVec(:,iTrj)))**2
```
marvinfriede, line 822:
```suggestion
                        rih = sum((xyz(:, nh) - xyz(:, i) - neigh%transVec(:, iTri))**2)
                        rjh = sum((xyz(:, nh) - xyz(:, j) + neigh%transVec(:, iTrj))**2)
```

src/gfnff/gfnff_ini2.f90
```diff
@@ -776,83 +776,107 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
-        enddo ! iTri
+         i=topo%hbatABl(1,ix)
+         j=topo%hbatABl(2,ix)
+         do iTri=1, neigh_nTrans ! go through i shifts
+            do iTrj=1, neigh_nTrans ! go through j shifts
+               ! get adjustet iTr -> for use of neigh% distances and bpair with two shifted atoms
+               iTrDum=neigh%fTrSum(neigh%iTrNeg(iTri),iTrj)
+               ! iTrDum=-1 is valid here because we are only interested if ij is a bond (ijnonbond)
+               ! However, iTrDum is not and should not be used as an index without excluding -1 value
+               if((iTrDum.le.neigh_nTrans.and.iTrDum.gt.0).or.iTrDum.eq.-1) then ! cycle invalid values
```
marvinfriede, line 806: Can you please convert these `if`s to early returns instead to make the code more readable?
I.e., invert the condition and use `continue`.

src/gfnff/gfnff_ini2.f90
```diff
@@ -776,83 +776,107 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
+                        free=.true. ! tripplet not assigned yet
+                        nh  =topo%hbatHl(1,k) ! nh always in central cell
+                        ! distances for non-cov bonded case
+                        rih=NORM2(xyz(:,nh)-(xyz(:,i)+neigh%transVec(:,iTri)))**2
+                        rjh=NORM2(xyz(:,nh)-(xyz(:,j)+neigh%transVec(:,iTrj)))**2
+                        ! check if i is the bonded A
+                        if(iTri.le.neigh_numctr) then ! nh is not shifted so bpair works without adjustment
+                           if(neigh%bpair(i,nh,iTri).eq.1.and.ijnonbond) then
+                              !$omp atomic capture
+                              nlist_nhb2 = nlist%nhb2
```
marvinfriede, line 827: Why did you introduce private variables instead of updating the shared variables under atomic protection?

src/gfnff/gfnff_ini2.f90
```diff
@@ -776,83 +776,107 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
       if(rmsd.lt.1.d-6 .or. rmsd.gt. 0.3d0) then ! update list if first call or substantial move occured
       nlist%nhb1=0
       nlist%nhb2=0
+      nlist_nhb1=0
+      nlist_nhb2=0
       ! loop over hb-relevant AB atoms
+      !$omp parallel do default(none) private(i, j, k, nh, ix, iTri, iTrj, iTrDum, rab, &
+      !$omp rih, rjh,ijnonbond,free ) &
+      !$omp firstprivate (nlist_nhb1, nlist_nhb2, neigh_numctr, neigh_nTrans, hbthr1, hbthr2) &
+      !$omp shared(nlist, topo, neigh, xyz)
```
marvinfriede, line 796: Maybe `schedule(dynamic)` improves performance here. The loop work could be different due to the conditions in the nested loops.

src/gfnff/gfnff_ini2.f90
Thomas3R: Good catch! Taking the square root to square it again is unnecessary. However, I think what we want here is the intrinsic `dot_product(vec, vec)` with `vec=xyz(:, nh) - (xyz(:, i) + neigh%transVec(:, iTri))`. Do you agree?

src/gfnff/gfnff_ini2.f90
Thomas3R: If I am not mistaken, continue is a "do-nothing" statement in Fortran. The previously implemented `cycle` statement is not allowed in OpenMP parallel do regions. Therefore, I changed it to the less readable if statement.

marvinfriede: Oh yes, I meant `cycle`. However, I do see why it should not be allowed. Actually, we use that rather often ([example](https://github.com/tblite/tblite/blob/d597263248ca5d64fba08accf17234cec5e141b7/src/tblite/repulsion/effective.f90#L159)).

src/gfnff/gfnff_ini2.f90
marvinfriede: @Albkat just mentioned that `dot_product` is a old Fortran function, and if one wants to use the dot product here BLAS level 1 functions should be used. However, I think just using sum and the square does the job.

src/gfnff/gfnff_ini2.f90
Thomas3R: You are absolutely right! Somehow my compiler told me I could not use `cycle` at the time, but it seems it has changed its mind about the current version of the code. I reversed it to the original version as requested.

src/gfnff/gfnff_ini2.f90
Thomas3R: Using the sum and square led to a change in energy, probably for numerical reasons. Using mctc_dot (which is the BLAS level 1 function) works as expected.

marvinfriede: The second line has a plus instead of a minus beforethe translation vector. That is probably the source of the error.

src/gfnff/gfnff_ini2.f90
foxtran: Better to use `collapse(3)` with evaluation of i/j variables in inner loops to increase OpenMP iteration space. Playing with second arg of dynamic scheduling may also help: I usually use 32 or 64.

src/gfnff/gfnff_ini2.f90
```diff
@@ -774,85 +775,110 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
+                     if(neigh%bpair(j,nh,iTrj).eq.1.and.ijnonbond) then
+                        !$omp atomic capture
+                        nlist%nhb2=nlist%nhb2+1
+                        nlist_nhb2 = nlist%nhb2
+                        !$omp end atomic
+                        nlist%hblist2(1,nlist_nhb2)=j
+                        nlist%hblist2(2,nlist_nhb2)=i
+                        nlist%hblist2(3,nlist_nhb2)=nh
+                        nlist%hblist2(4,nlist_nhb2)=iTrj
+                        nlist%hblist2(5,nlist_nhb2)=iTri
```
foxtran, line 853: Here you may have data race with nlist_nhb2 variable. On x86/x86_64 it works fine, but AArch64 can say "Hi" for you :)

src/gfnff/gfnff_ini2.f90
```diff
@@ -774,85 +775,110 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
+      !$omp end parallel do
 
       ! for nxb list only i is not shifted
       nlist%nxb =0
       do ix=1,topo%natxbAB
           i =topo%xbatABl(1,ix)   ! A
           j =topo%xbatABl(2,ix)   ! B
           iTrj=topo%xbatABl(4,ix) ! iTrB
-          if(iTrj.gt.neigh%nTrans.or.iTrj.lt.-1.or.iTrj.eq.0) cycle ! cycle nonsense 
+          if(iTrj.gt.neigh_nTrans.or.iTrj.lt.-1.or.iTrj.eq.0) cycle ! cycle nonsense
```
foxtran, line 881: Why was it changed?

src/gfnff/gfnff_ini2.f90
```diff
@@ -774,85 +775,110 @@ subroutine gfnff_hbset(n,at,xyz,topo,neigh,nlist,hbthr1,hbthr2)
...
+               endif
+               ! loop over relevant H atoms
+               do k=1,topo%nathbH
+                  free=.true. ! tripplet not assigned yet
+                  nh  =topo%hbatHl(1,k) ! nh always in central cell
+                  ! distances for non-cov bonded case
+                  vec_ih = xyz(:,nh)-(xyz(:,i)+neigh%transVec(:,iTri))
+                  rih = mctc_dot(vec_ih, vec_ih)  ! square of distance iH
+                  vec_jh = xyz(:,nh)-(xyz(:,j)+neigh%transVec(:,iTrj))
+                  rjh = mctc_dot(vec_jh, vec_jh)  ! square of distance jH
```
foxtran, line 826: It is better to use `dot_product` intrinsic (which can be optimized by compiler) rather than using external procedure mctc_dot

PR #1244, opened 2025-03-29
Licen -> License

foxtran:
By some reason (most probably F77->F90 conversion), all lines were cut to 72 symbols, while some strings had more than 72 symbols. This patch fixes those lines.


marcelmbn (approved):
LGTM! (Only changes in comments anyway)

PR #1178, opened 2025-02-05
Speed-up hot path of GFN2 geometry optimization 

foxtran:
This PR improves parallelization for GFN2 geometry optimization. It affects other types of calculations too.

Achieved speed-up is **30%**. Gradients are **2x** faster!

For 578 atom structure, before the patch:
```
 total:
 * wall-time:     0 d,  0 h,  1 min, 11.680 sec
 *  cpu-time:     0 d,  0 h,  3 min,  7.524 sec
 * ratio c/w:     2.616 speedup
 SCF:
 * wall-time:     0 d,  0 h,  0 min, 13.534 sec
 *  cpu-time:     0 d,  0 h,  0 min, 41.212 sec
 * ratio c/w:     3.045 speedup
 ANC optimizer:
 * wall-time:     0 d,  0 h,  0 min, 57.390 sec
 *  cpu-time:     0 d,  0 h,  2 min, 25.103 sec
 * ratio c/w:     2.528 speedup

normal termination of xtb

real 1m11.740s
user 3m5.985s
sys 0m1.572s
```

After the patch:
```
 total:
 * wall-time:     0 d,  0 h,  0 min, 48.779 sec
 *  cpu-time:     0 d,  0 h,  2 min, 51.122 sec
 * ratio c/w:     3.508 speedup
 SCF:
 * wall-time:     0 d,  0 h,  0 min, 11.737 sec
 *  cpu-time:     0 d,  0 h,  0 min, 42.249 sec
 * ratio c/w:     3.600 speedup
 ANC optimizer:
 * wall-time:     0 d,  0 h,  0 min, 36.377 sec
 *  cpu-time:     0 d,  0 h,  2 min,  8.025 sec
 * ratio c/w:     3.519 speedup

normal termination of xtb

real 0m48.971s
user 2m50.138s
sys 0m1.012s
```

Command for testing:
```
xtb --input model.xtb model.xyz -P 4 --gfn 2 --opt vtight --alpb water --cycles 5
```

model.xtb (nothing interesing, to be honest):
```
$opt
    engine=rf
$end
$alpb
    kernel=still
    grid=tight
$end
```

Build type: CMake, Release, gfortran-14, Intel MKL, no march (can get extra speed-up), enabled debug info.

PR #1253, opened 2025-04-04
Speed-up gradient evaluation

foxtran:
This PR introduces a small set of OpenMP optimizations which slightly speeds up each geometry optimization step.

Tested with ifx 2025.1.

See commit messages about impact of each of optimizations.

### DIFF
--- a/src/disp/dftd4.F90
+++ b/src/disp/dftd4.F90
@@ -688,7 +688,8 @@ subroutine build_wdispmat(dispm,nat,ndim,at,itbl,xyz,par,c6abns,gw,wdispmat)
    !$omp parallel do shared(wdispmat) &
    !$omp shared(nat, at, xyz, itbl, par, dispm, c6abns, gw) &
    !$omp private(ia, k, j, ja, l, r4r2ij, cutoff, r2, oor6, oor8, oor10, &
-   !$omp& ii, jj, gwgw, c8abns, c10abns)
+   !$omp& ii, jj, gwgw, c8abns, c10abns) &
+   !$omp schedule(dynamic,32) collapse(2)
 !#endif
    do i = 1, nat
       do j = 1, nat
@@ -1104,7 +1105,8 @@ subroutine get_atomic_c6(dispm, nat, atoms, zetavec, zetadcn, zetadq, &
    !$omp parallel do default(none) shared(c6, dc6dcn, dc6dq) &
    !$omp shared(nat, atoms, dispm, zetavec, zetadcn, zetadq) &
    !$omp private(iat, ati, jat, atj, dc6, dc6dcni, dc6dcnj, dc6dqi, dc6dqj, &
-   !$omp& iref, jref, refc6)
+   !$omp& iref, jref, refc6) &
+   !$omp schedule(dynamic,32) collapse(2)
 #endif
    do iat = 1, nat
       do jat = 1, nat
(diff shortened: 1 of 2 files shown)


foxtran:
Ah... It is overlapped with #1178

marvinfriede:
I did some testing with different scheduling settings in DFT-D4. I found that `dynamic,1` (and `guided`, which probably just uses `dynamic,1`) are the best choices, because the loops have considerably different lengths. I did not see the specific chunk size to be more effective than `dynamic,1`. Moreover, I also tested the original triangular iteration space and the rectangular (collapsed) iteration space and did only see minor differences with `dynamic,1`; in fact, the collapsed version was slightly slower.
I assume you see these speed-ups, because you switch from `static` to `dynamic`. The default when compiling with `ifx` is `static,1`, which is the worst case for the nested loops of different lengths we have: We parallelize the first loop over `i` and the second, unparallelized loop over `j` only goes up to `i`, i.e., for small `i` the work is small because the `j` loop also terminates quickly, but for large `i`, `j` is also large. This gets worse for three nested loops and the gains of `dynamic` are even bigger (I found a factor of 2 for the ATM term in D4 by switching from defaults, i.e. `static`, to `dynamic` or `guided`).
I do not think collapsing really helps here, and neither does the specific chunk size.

marvinfriede:
> For which of cycle?

Do you mean which loops? [Two-body dispersion term](https://github.com/dftd3/simple-dftd3/blob/be6c011c03fe976060ef9af9056d57547d8cbf3d/src/dftd3/damping/rational.f90#L170) and [three-body ATM dispersion term](https://github.com/dftd3/simple-dftd3/blob/be6c011c03fe976060ef9af9056d57547d8cbf3d/src/dftd3/damping/atm.f90#L128).

Correction to the previous comment: I tested in D3 (s-dftd3) to eliminate the time for solving the linear system in D4's charge model.

foxtran:
I have played with [three-body ATM dispersion term](https://github.com/dftd3/simple-dftd3/blob/be6c011c03fe976060ef9af9056d57547d8cbf3d/src/dftd3/damping/atm.f90#L128) and `(dynamic,1)` saves about 30 ms for 600 atomic systems in comparison with `collapse(2) schedule(dynamic,32)` or ~7-8 % of comput.time. Simple `(dynamic,32)` spends extra 20 ms in comparison with collapsed version.

foxtran:
And for large number of cores `collapse(2)` works much better :)

@marvinfriede, what is the target CPU cores number?

marvinfriede:
I only tested up to 12 cores. Did you test more and saw that `collapse(2)` is better?
What about `guided` for more cores?

foxtran:
> What about guided for more cores?

It requires a lot of syncs between cores, so I just said no for this kind of schedule :)

>  Did you test more and saw that collapse(2) is better?

For single AMD 7354, `collapse(2)` became better from 9 cores.

marvinfriede:
> It requires a lot of syncs between cores, so I just said no for this kind of schedule :)

I did not observe much difference to `dynamic,1`, and it was always one of the best performing scheduling settings.
 
> For single AMD 7354, `collapse(2)` became better from 9 cores but it looks like that depends on input.

So, there is no clear evidence that `collapse(2)` should be preferred over the triangular iteration space?

foxtran:
> So, there is no clear evidence that collapse(2) should be preferred over the triangular iteration space?

Yes. Anyway, I found a way to make it 1/3 faster again :D

foxtran:
There is a small snippet which now works properly. Unfortunately, after its debugging, it saves only few milliseconds and therefore impact of CPU freq is higher that from this change and I am not able to measure timings properly. 

```Fortran
   !>
   !> @brief get `i`, `j` indexes of matrix from triangle index `t`
   !>
   !> @details to avoid heavy sqrt function, function guesses row index several times 
   !>
   !> @param[in]    t - triangle index
   !> @param[inout] i - row index
   !> @param[out]   j - col index
   pure subroutine tij(t, i, j)
      integer, intent(in) :: t
      integer, intent(inout) :: i
      integer, intent(out) :: j
      integer :: c
      ! guess that we already have a proper `i`
      j = t - (i-1) * i / 2
      ! some schedulers may have non-monotonic iterations
      ! for them, we go directly to evaluation of `i` without guessing
      if (j <= 0) goto 10
      ! col index is OK; we can continue calculation 
      if (j <= i) return
      ! do several attemptions...
      do c = 1, 3
         j = j - i
         i = i + 1
         if (j <= i) return
      end do
      ! we failed; time for math
   10 continue
      i = (int(sqrt(real(1 + 8 * (t-1), kind=4))) - 1) / 2 + 1
      j = t - (i-1) * i / 2
   end subroutine tij
```

With it, one can use expanded triangle space for iterations. So, instead:
```
!$omp collapse(2)
do iat = 1, nat
do jat = 1, nat
if (jat > iat) cycle
```
one can write:
```
!$omp ...
do tat = 1, nat*(nat+1)/2
call tij(tat, iat, jat)
```

foxtran:
The snippet may help for CPU if sqrt evaluation is quite long. Modern x86_64 processors can do it in less than 20 cycles and for them snippet does not give any improvements.

thfroitzheim:
What is the status here? Is this the final parallelization or are you still working on it? Given the extensive testing also in #1272 this looks fine to me

foxtran:
> Is this the final parallelization or are you still working on it?

It is final; most of changes from this PR were merged in #1178

PR #1268, opened 2025-04-15
Avoid infinite `mopac` loops 

### LINKED ISSUE #1267: `mopac` calculation gets stuck in infinite loop for single atoms
**Describe the bug**
`mopac` calculation gets stuck in infinite loop for single atoms.

**To Reproduce**
PM6-D3H4X calculation for:
```
1

F 0.0 0.0 0.0
```

**Expected behaviour**
Calculation should run through and not get stuck reading an output file forever.

marcelmbn:
- implemented some exits on fail checks
- we don't expect gradients for atoms
- closes #1267


thfroitzheim (approved):
LGTM

Issue #1258, opened 2025-04-09
xTB inputs for benchmarking


foxtran:
Is there any good inputs for checking xTB speed?

foxtran:
There is an archive with some structures from 30 to 6000 atoms. 

0090_0030xWATER.xyz was taken from https://ravel.pctc.uni-kiel.de/research-topics/ttm2-f-water-clusters/
0417_0139xWATER.xyz	0570_0190xWATER.xyz, 0711_0237xWATER.xyz, 0909_0303xWATER.xyz, 1707_0569xWATER.xyz were taken from http://www.ergoscf.org/xyz/h2o.php
GluAla.xyz's were taken from http://www.ergoscf.org/xyz/gluala.php

[bench.xyz.zip](https://github.com/user-attachments/files/19824446/bench.xyz.zip)

PR #1272, opened 2025-04-23
Small speed-up of gradient for non-PBC case

foxtran:
I have split PBC and non-PBC case in `atm_gradient_latp` subroutine that saves 30 ms for `0418_0016xGluAla.xyz` (418 atoms) and 300 ms for `0834_0032xGluAla.xyz` (834 atoms) per each geom.iter for 4 cores with ifx 2025.1. For larger number of atoms the effect should be larger.


src/disp/dftd4.F90
```diff
@@ -2038,71 +2038,132 @@ subroutine atm_gradient_latp &
...
             end do
+         end do
+      end do
+      !$omp end parallel do
+   else
+      !$omp parallel do default(none) reduction(+:energies, gradient, sigma, dEdcn) &
+      !$omp shared(mol, r4r2, par, trans, cutoff2, c6, dc6dcn, nat) &
+      !$omp private(iat, ati, jat, atj, kat, atk, c6ij, cij, c6ik, c6jk, cik, cjk, &
+      !$omp& rij, r2ij, rik, r2ik, rjk, r2jk, scale, dE, dG, dS, dCN) &
+      !$omp collapse(2) schedule(dynamic,32)
```
thfroitzheim, line 2116: As mentioned in #1253, the chunk size of 32 is not ideal for the ATM term since the loops are maximally unbalanced. Can you also please check whether the collapse is actually beneficial here. From our tests it didn't help much and dynamic,1 without collapse was ideal

src/disp/dftd4.F90
```diff
@@ -2038,71 +2038,132 @@ subroutine atm_gradient_latp &
...
-                     & - trans(:, jtr)
-                  r2jk = sum(rjk**2)
-                  if (r2jk > cutoff2 .or. r2jk < 1.0e-14_wp) cycle
-
-                  call deriv_atm_triple(c6ij, c6ik, c6jk, cij, cjk, cik, &
-                     & r2ij, r2jk, r2ik, dc6dcn(iat,jat), dc6dcn(jat,iat), &
-                     & dc6dcn(jat,kat), dc6dcn(kat,jat), dc6dcn(iat,kat), &
-                     & dc6dcn(kat,iat), rij, rjk, rik, par%alp, dE, dG, dS, dCN)
+   doPBC = .false.
+   if (size(trans, dim=2) > 1) doPBC = .true.
```
thfroitzheim, line 2047: Seems to be a modest speed up for a lot of duplicate code. Is this a relevant fraction of the compute time for fast methods such as the force field?

thfroitzheim (changes requested):

src/disp/dftd4.F90
foxtran: > the chunk size of 32 is not ideal for the ATM term since the loops are maximally unbalanced.

Unfortunately, it strongly depends on used compiler / OpenMP runtime. For intel runtime, `(dynamic,1)` is fine, that  is true. 

I have to find a machine with locked CPU frequency where I can test it really well. Currently, I'm waiting couple minutes before seeing the difference between dynamic,1 / dynamic,32 for intel runtime.

> From our tests it didn't help much

Could you please share them in #1258?

src/disp/dftd4.F90
foxtran: It is even relevant for xTB-GFN2: for us it saves about 8-12% of total CPU time (2.5s SCC + 1.0 grad -> 2.5s SCC + 0.8s grad).

src/disp/dftd4.F90
foxtran: Some internal testing (610 atoms)

Intel LLVM Compiler, ifx 2025.1.0:
(dynamic, 1), 6 threads: 680.9 +/- 3.0 ms
(dynamic, 32), 6 threads: 679.7 +/- 3.8 ms
(dynamic, 64), 6 threads: 678.5 +/- 3.1 ms

Here, I have measured only timing related to this OMP section. So, as you can see, there is no big difference between different batch sizes.

foxtran: Data from #1258: bench.xyz/0418_0016xGluAla.xyz
Command:
`../xtb bench.xyz/0418_0016xGluAla.xyz -P 4 --gfn 2 --opt --cycles 10`

gfortran 14.2, JH7110@1.5GHz:
(dynamic, 1), 4 threads: 4.955s, 4.957s
(dynamic, 32), 4 threads: 4.952s, 4.948s

As you can see, there is few ms difference between (dynamic, 1) and (dynamic, 32), but the second one is faster.

thfroitzheim: Ok, those changes are generally quite small for the slightly more confusing code, but that is not really much of a difference. Is the if statement in the loop also fine without omp parallelization? If this is the case, we can switch to this loop layout

foxtran: > Is the if statement in the loop also fine without omp parallelization?

Compiler should understand it well. See line 34 in asm here: https://godbolt.org/z/8Ts6Keo3n

thfroitzheim (approved):

PR #1281, opened 2025-05-12
Avoid spread intrinsic

foxtran:
With gfortran, spread intrinsic is not optimized (more precisely this is libgfortran call) and therefore it significantly affects on performance, especially for 3-body terms. 

This patch speeds up gradient code in 1.5-2x times with gfortran as well other parts which I did not measure. By some reason, it also affects ifx but it gives only 10% speed up for `deriv_atm_triple` subroutine. 

I have used unrolled cycles since they provide an extra optimizations for ifx: time was improved from 680 to 625 ms for my input. Usual cycles does not give this improvement: see the assembler difference here: https://godbolt.org/z/zs6e7vzT4. The first source code is the original, the second is presented in this PR and the third one uses cycles. 

I did not touched initializations/io parts where spread are used.

### DIFF
--- a/src/coulomb/gaussian.f90
+++ b/src/coulomb/gaussian.f90
@@ -432,7 +432,9 @@ subroutine getCoulombDerivsCluster(mol, itbl, rad, qvec, djdr, djdtr, djdL)
                gij = 1.0_wp/(rad(ish, iid)**2 + rad(jsh, jid)**2)
                g1 = erf(sqrt(gij*r2))/sqrt(r2)
                dG(:) = (2*sqrt(gij)*exp(-gij*r2)/sqrtpi - g1) * vec/r2
-               dS(:, :) = 0.5_wp * spread(dG, 1, 3) * spread(vec, 2, 3)
+               dS(:, 1) = 0.5_wp * dG(1) * vec
+               dS(:, 2) = 0.5_wp * dG(2) * vec
+               dS(:, 3) = 0.5_wp * dG(3) * vec
                djdr(:, iat, jj+jsh) = djdr(:, iat, jj+jsh) - dG*qvec(ii+ish)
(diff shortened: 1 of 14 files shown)


src/coulomb/gaussian.f90
```diff
@@ -432,7 +432,9 @@ subroutine getCoulombDerivsCluster(mol, itbl, rad, qvec, djdr, djdtr, djdL)
                gij = 1.0_wp/(rad(ish, iid)**2 + rad(jsh, jid)**2)
                g1 = erf(sqrt(gij*r2))/sqrt(r2)
                dG(:) = (2*sqrt(gij)*exp(-gij*r2)/sqrtpi - g1) * vec/r2
-               dS(:, :) = 0.5_wp * spread(dG, 1, 3) * spread(vec, 2, 3)
+               dS(:, 1) = 0.5_wp * dG(1) * vec(:)
```
thfroitzheim, line 435:
```suggestion
               dS(:, 1) = 0.5_wp * dG(1) * vec
```
They shouldn't be necessary here and in the following

thfroitzheim:
> I did not touch spread hell in `gfnff/gfnff.f90`. Also, I did not touched initializations/io parts where spread also used.

I know this is the annoying part, but GFN-FF would likely benefit the most from improved parallelization (i.e. in #1240). Likely much more than just a few ms here and there

foxtran:
> Likely much more than just a few ms here and there

It is not a few ms for here and here for GCC, unfortunately.

src/coulomb/gaussian.f90
foxtran: Fixed.

foxtran:
For `bench.xyz/1666_0064xGluAla.xyz`, for GCC I got a 20% speed-up with this patch: from 0.78 sec/iter to 0.63 sec/iter in single-threaded run.

thfroitzheim (approved):
LGTM

PR #1292, opened 2025-05-21
Fix copying of TMolecule

foxtran:
By some reason, GCC 12.2 does not copy some data between TMolecule objects and, therefore, there is some trash in output like (it is xtbopt.xyz):
image

This patch fixes it. Not sure that I found all places, but at least GFN2 geometry optimizations work well.


src/type/molecule.f90
```diff
@@ -198,8 +198,10 @@ subroutine copyMolecule(self,mol0)
    class(TMolecule), intent(out) :: self
    type(TMolecule), intent(in) :: mol0
 
-      Call init(self, mol0%at, mol0%sym, mol0%xyz, mol0%chrg, mol0%uhf, &
+   call self%deallocate()
+   Call init(self, mol0%at, mol0%sym, mol0%xyz, mol0%chrg, mol0%uhf, &
            & mol0%lattice, mol0%pbc)
+   self%ftype = mol0%ftype
 end subroutine copyMolecule
```
thfroitzheim, line 208: Maybe we should merge the TMolecule copy part here with the `=` assignment and add a deep copy of everything. Then one could still just use the assignment with `=`

src/optimizer.f90
TyBalduf, line 596: I believe this one also needs to be updated to `call mol%copy(molopt)` or else the gfortran issues persist, as reported in https://github.com/grimme-lab/xtb/issues/1391 and https://github.com/grimme-lab/xtb/issues/1380

thfroitzheim:
```suggestion
   call mol%copy(molopt)
```

I agree this should also be replaced

thfroitzheim (approved):
Let's use this fix even before merging everything the `=` assignment

Issue #1308, opened 2025-06-10
Error on optimization using ORCA gradients (ORCA 6+)


lmseidler:
**Describe the bug**
When trying to run a geometry optimization using the ANCOPT driver using ORCA gradients (! ENGRAD) xtb fails to run due to a check for the `orca_scf` executable in the ORCA directory.

**To Reproduce**
Steps to reproduce the behaviour:

1. have ORCA 6+ set up such that path starts with `/usr/`
2. take arbitrary input, set up xcontrol for ANCOPT using ORCA
3. prepare ORCA input including ENGRAD
4. run `xtb <geometry input> --opt normal --orca -I <xcontrol file>`

The error does not occur if the ORCA path does not start with `/usr/`.

**Expected behaviour**
xtb should run the calculation.

**Additional context**
-

PR #1309, opened 2025-06-10
ORCA executable check vs gnome screen reader using SCF keyword in binary

lmseidler:
Fixes a bug as mentioned in #1308 with a better sanity check. This now checks the orca binary contents for the keyword "SCF" as this should be specific enough to discriminate it from the GNOME screen reader binary.
Also formatted the source file.

### DIFF
--- a/src/extern/orca.f90
+++ b/src/extern/orca.f90
@@ -17,16 +17,16 @@
 
 module xtb_extern_orca
 !$ use omp_lib
-   use xtb_mctc_accuracy, only : wp
-   use xtb_mctc_filetypes, only : fileType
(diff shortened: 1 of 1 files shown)


awvwgk (commented):
How about checking for `orca_2json` binary or similar and reconstructing the path to the main Orca binary from it?

lmseidler:
Since the user provides the path to the orca binary I did not consider trying to automatically detect the binary. Also, checking for a specific binary might not be version agnostic, this is the reason the check broke in the first place. I confirmed that checking for 'SCF' works for ORCA 4-6.

Albkat (approved):

PR #1316, opened 2025-06-22
Write output JSON after frequency analysis

AlbertoBoldrini:
Previously, the thermodynamic properties were not written to the JSON output due to the incorrect order of operations.


thfroitzheim (approved):
This order really didn't make much sense before. Thank you for this contribution!

PR #1324, opened 2025-07-02
Update build system

foxtran:
I have fixed fetching HEAD instead of specific revision. Also, IFX 2025.2 is used now in CI.


awvwgk (approved):

toxtran:
@awvwgk, ping :-)

thfroitzheim:
I don't really want to merge a new failing test. Since it seems to be a problem with the combination of macos-latest and gcc-13, we might want to exclude this combination (maybe replace it with an older macos version and gcc-13). Otherwise, we don't know whether future PRs actually break something.

toxtran:
> Otherwise, we don't know whether future PRs actually break something.

You can always check the status of separate CI runs. I do not see any problems here.

> problem with the combination of macos-latest and gcc-13,

Just fix the macos + gcc-13 in the next PR :-) 
In my opinion, it would be better to do in separate PRs.

Or I will fix it once I fix my eye. It does not look difficult.

Issue #1159, opened 2025-01-14
ddCOSMO not variational


lukaswittmann:
**Describe the bug**
ddCOSMO energy is not variational. Thus, this also applies to CPCM-X.
This also applies to tblite and its cpcm.

**To Reproduce**
Easy to reproduce using NaCl and water:
```
2

Na 0 0 0
Cl 0 0 1.5
```
via `xtb nacl.xyz --cosmo water`

Output during SCC:
```
 iter      E             dE          RMSdq      gap      omega  full diag
   1     -4.8764669 -0.487647E+01  0.386E+00    5.88       0.0  T
   2     -4.8820210 -0.555413E-02  0.234E+00    5.26       1.0  T
   3     -4.8900739 -0.805288E-02  0.281E-01    4.37       1.0  T
   4     -4.8893756  0.698224E-03  0.186E-01    4.48       1.0  T
   5     -4.8891790  0.196604E-03  0.100E-01    4.52       1.0  T
   6     -4.8895562 -0.377156E-03  0.302E-02    4.48       2.3  T
   7     -4.8895093  0.468651E-04  0.953E-03    4.48       7.4  T
   8     -4.8894764  0.329602E-04  0.825E-04    4.48      85.7  T
   9     -4.8894765 -0.155453E-06  0.390E-05    4.48    1813.1  T
  10     -4.8894765  0.287508E-07  0.974E-06    4.48    7258.5  T
```
with the final energy:
```
         :: SCC energy                -4.889476488844 Eh    ::
```
The final energy is not the lowest energy.

Issue #1163, opened 2025-01-30
ddCOSMO Surface Area Broken


lukaswittmann:
The surface area written into the `xtb.cosmo` file is wrong for any overlapping spheres.

Example: Surface area (Ang^2) of O2 from distance 0 to 4 Angs in 0.01 Ang steps (default):
![Image](https://github.com/user-attachments/assets/8853ce8f-e030-4a17-b5ea-6655220e60ad)

It is discontinuous and qualitatively wrong. One would expect a nearly linear growth from `a` to `2a`. This is correct, e.g., in ALPB.[
The respective routine](https://github.com/grimme-lab/xtb/blob/0fb151dceb19b8df9f6012789cf7d2dd6e4a6cd2/src/solv/cosmo.f90#L875-L876) misses the inclusion of the switching function `ui` (via `self%ddCosmo%ui(ig, iat)`); with inclusion (swi), however, the surface area is still incorrect.

The respective implementation in our [tblite ddx development branch](https://github.com/lukaswittmann/tblite) delivers the correct result.

lukaswittmann:
Small addon: This is the same surface area calculated with numsa (uses SwiG discretization):

![Image](https://github.com/user-attachments/assets/ae5f45af-369d-4482-9dcc-1d0c7f25f80f)

PR #1325, opened 2025-07-02
Remove CPCM-X library

foxtran:
This PR removes dependency of [CPCM-X library](https://github.com/grimme-lab/CPCM-X).

The reasons why:
- often leads to failure of CI tests that may affect users too
- no geometry optimizations
- API is used through files

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1742,6 +1742,9 @@ contains
                call env%error("No solvent name provided for COSMO", source)
             end if
 
+         case ('--cpcmx')
+            call env%error("The CPCM-X library was excluded starting from xtb 6.7.1. Please, use older xtb.", source)
+
          case ('--scc', '--sp')
             call set_runtyp('scc')
 


thfroitzheim:
While I see the stability problems with CPCM-X, it is an actively used feature of xtb and cannot just be removed. We are currently working on new solvation models, so maybe we can replace CPCM-X at some point in the future

src/features.F90
```diff
@@ -42,8 +37,6 @@ function get_xtb_feature(feature) result(has_feature)
    select case(feature)
    case("tblite")
       has_feature = tblite_support
-   case("cpcmx")
-      has_feature = cpcmx_support
```
awvwgk, line 46: Let's keep this and always return false here

src/prog/main.F90
```diff
@@ -1767,20 +1738,6 @@ subroutine parseArguments(env, args, inputFile, paramFile, lgrad, &
                call env%error("No solvent name provided for COSMO", source)
             end if
 
-         case ('--cpcmx')
```
awvwgk, line 1774: Maybe keep this and make it an error

awvwgk (approved):
Agree that this might be the right way forward for now. If this feature will be reactivated we can bring it back again from the git history.

src/features.F90
foxtran: Ok. Give me couple minutes ;-)

foxtran: cpcmx will go to `case default`. So, no changes here.

src/prog/main.F90
foxtran: Fixed.

corinwagen:
We'd love to help but neither @jevandezande nor I know FORTRAN, sadly

lukaswittmann:
The problem with implementing this in, e.g., tblite will be, that we fixed #1159 (and #1163). This means, that CPCM-X will not work with the new implementation (i.e., the results will likely be not usable without a rework and parameterization) and it would require us to ship two different implementations of the ddCOSMO scheme.

PR #1327, opened 2025-07-10
electronic temperature now also set for tblite correctly via command line flag --etemp

gorges97:
If the electronic temperature was specified via the --etemp command line flagcommand, its value did not change when using --tblite.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1614,6 +1614,8 @@ contains
             call args%nextArg(sec)
             if (allocated(sec)) then
                call set_scc(env, 'temp', sec)
+               !set etemp for tblite
+               idum = getValue(env, sec, tblite%etemp)
             else
                call env%error("Temperature in --etemp option is missing", source)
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Good catch, thanks for fixing

src/prog/main.F90
```diff
@@ -1614,6 +1614,8 @@ subroutine parseArguments(env, args, inputFile, paramFile, lgrad, &
             call args%nextArg(sec)
             if (allocated(sec)) then
                call set_scc(env, 'temp', sec)
+               !set etemp for tblite
+               idum = getValue(env, sec, tblite%etemp)
```
foxtran, line 1618: This patch added the following warning:

```
main.F90:1627:22:

 1627 |                idum = getValue(env, sec, tblite%etemp)
      |                      1
Warning: Extension: Conversion from LOGICAL(4) to INTEGER(4) at (1)
```

PR #1331, opened 2025-07-31
Use dummy logical variable

foxtran:
This PR removes warning introduced in #1327


toxtran:
@thfroitzheim, could you please merge it?

thfroitzheim:
But `etemp` is not a logical but a real (i.e., it is the electronic temperature, which should just be a number).

PR #1344, opened 2025-09-27
Bump mctc-lib to v0.5.0

foxtran:
It helps to compile xtb with LLVM Flang.

xtb + OpenMP + tblite works well with LLVM Flang 21.1.1. Some patches of tblite are necessary for compiling which are absent in v0.4.0.


foxtran:
> Is the issue with the macos test here also the same as in https://github.com/grimme-lab/xtb/pull/1292?

I have no idea.

foxtran:
Looks like it is better to test mctc with all combinations and return here again.

PR #1348, opened 2025-11-07
docs: add --cycles to man page

nmaeder:
Adds a description for the --cycles parameter to the man page.

The parameters is described in the documentation [read the docs](https://xtb-docs.readthedocs.io/en/latest/optimization.html) and in the [xtb-docs github](https://github.com/grimme-lab/xtb_docs/blob/main/source/optimization.rst) but does not occur in the man page.


nmaeder:
i will also quickly add it to the help subroutine so it also shows there, fine with you @awvwgk ?

nmaeder:
Additionally added it to both the help and help_legacy subroutine.

thfroitzheim (approved):

PR #1351, opened 2025-11-07
Fix CI

foxtran:
More tests should be green now

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -20,8 +20,6 @@ jobs:
                - os: macos-15-intel
                  version: 15
                - os: macos-latest
-                 version: 13
-               - os: macos-latest
                  version: 14
                - os: macos-latest
                  version: 15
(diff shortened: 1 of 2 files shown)


thfroitzheim:
Given that macos-13 runners will be [discontinued](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/) anyway, I think this update is a good idea. Since we would loose support for `x86_64` intel versions of macos, we could try whether the new `macos-15-intel` runner works out of the box. If not, I would be fine with dropping the support for the intel MacOS

foxtran:
In general, CI is more green now. gcc-13 + meson + macos is always failing. gcc-14 + cmake + ubuntu is flipping :(

thfroitzheim:
Then let's exclude the gcc-13 + macos combination. We saw this already before that there is something strange going on. newer gcc versions seem to work fine. gcc-14 + cmake + ubuntu seems to be the old gfn0 issue. I would just keep this in the CI (this seems to be a real numerical problem and not just some compiler specific stuff).

Issue #1062, opened 2024-07-04
Add libiomp5md.dll to xtb-6.7.0-Windows-x86_64.zip


mdtoo:
Hello

Previous versions of xtb for the Windows binaries (6.5.0, 6.5.1, 6.6.0, 6.6.1) comes equipped with libiomp5dmd.dll. However, the latest version of 6.7.0 does not, requiring the user to download a previous version of xtb to obtain the necessary file.

Adding libiomp5dmd.dll directly to the latest Windows distribution would be much appreciated.

PR #1354, opened 2025-11-08
Add static windows compile to workflow and continuous delivery

thfroitzheim:
This PR adds a static windows compile and packing with the necessary DLLs to the Github workflow (this should address #1062. Due to unavailability of the ifort compilers on Windows and problems with ifx, I moved it to gfortran-15. I also added the binary to the continuous delivery. 

This should then also allow for a new release with a Windows binary. @Albkat maybe we can test on a Windows machine if the binary actually works (I can only test the testsuite here).

### DIFF
--- a/.github/workflows/fortran-build.yml
+++ b/.github/workflows/fortran-build.yml
@@ -388,7 +388,6 @@ jobs:
             install: >-
               git
               mingw-w64-${{ matrix.arch }}-gcc-fortran
-              mingw-w64-${{ matrix.arch }}-openblas
               mingw-w64-${{ matrix.arch }}-lapack
               mingw-w64-${{ matrix.arch }}-python
               mingw-w64-${{ matrix.arch }}-python-pip


thfroitzheim:
The continuous delivery also works with this update (see [here](https://github.com/thfroitzheim/xtb/actions/runs/19195935835/job/54877205161)). However, I avoid triggering it for a PR.

foxtran:
> The Intel/ifx build works also but leads to small (10^-7) deviations in some tests

Could  you please create bug reports about this?

.github/workflows/fortran-build.yml
```diff
@@ -180,7 +184,14 @@ jobs:
             msystem: ${{ matrix.msystem }}
             update: false
             install: >-
-               git mingw-w64-${{ matrix.arch }}-gcc-fortran mingw-w64-${{ matrix.arch }}-openblas mingw-w64-${{ matrix.arch }}-lapack mingw-w64-${{ matrix.arch }}-python mingw-w64-${{ matrix.arch }}-python-pip mingw-w64-${{ matrix.arch }}-meson mingw-w64-${{ matrix.arch }}-ninja
+               git
+               mingw-w64-${{ matrix.arch }}-gcc-fortran
+               mingw-w64-${{ matrix.arch }}-openblas
```
Albkat, line 189: Does one really need OpenBLAS, if we are linking to Netlib?

.github/workflows/fortran-build.yml
```diff
@@ -326,25 +351,118 @@ jobs:
...
          - name: Run unit tests
-           run: >-
-            meson test -C ${{ env.BUILD_DIR }} --print-errorlogs --num-processes 1 --no-rebuild --suite xtb -t 120
+           run: >
+            meson test -C ${{ env.BUILD_DIR }} --print-errorlogs --num-processes 1
+            --no-rebuild --suite xtb -t 120
            env:
             OMP_NUM_THREADS: 2,1
+
+   mingw-meson-static-build:
```
Albkat, line 370: Really nice work, building it without the actual Windows machine ;)

.github/workflows/fortran-build.yml
thfroitzheim: No, this was just a relic of earlier attempts.

Issue #1334, opened 2025-08-06
xTB DIPRO - orbital overlap units


ricalessandri:
I've been trying out the xTB DIPRO and was wondering why the orbital overlap appears in Eh units. Shouldn't it be unit-less, or am I missing something.

This can be seen in the [sample output in the documentation](https://xtb-docs.readthedocs.io/en/latest/dipro.html):
```
Fragment 1 HOMO   Fragment 2 HOMO
--------------------------------------
E_mon(orb) frag1 frag2             -12.966   -12.099 eV
|J(AB)|:                0.238 eV
S(AB):            -0.00998928 Eh
|J(AB,eff)|:            0.113 eV
```

thfroitzheim:
Yes, the overlap shouldn't have a dimension. Should be an easy fix here: [src/dipro.F90#L414](https://github.com/grimme-lab/xtb/blob/902b313678b95d793122174df09d590365a669d7/src/dipro.F90#L414)

PR #1356, opened 2025-11-10
fix printout of dipro

chselz:
should fix issue #1334 

i tested the built with ifort 2021.6.0 and the meson build system


thfroitzheim:
Please correct the output also in the [documentation](https://xtb-docs.readthedocs.io/en/latest/dipro.html)

PR #1358, opened 2025-11-19
write current iter in xtbopt.log commentline

### LINKED ISSUE #411: Include optimization cycle in optimization log
Write the optimization step in the comment line of the optimization log

The `writeMolecule` already supports a `number`/`cycle` argument which currently only used for PDB but could be extended for other formats as well

chselz:
I modified the `writeMolecule` subroutine to use the previously unused number argument to write out the current iteration in the geometry optimization. I implemented and tested this for ANCopt L-ANCopt, and FIRE optimization engines. For periodic boundary conditions, it was apparently already done in the xtboptlog.cif file.

To do this, I had to shift some statements so that the iter variable is initialized before using it in the `writeMolecule` statement. 

I built it with meson and ifort.

This should fix #411

### DIFF
--- a/src/io/writer.f90
+++ b/src/io/writer.f90
@@ -51,6 +51,7 @@ subroutine writeMolecule(self, unit, format, energy, gnorm, number)
    character(len=:), allocatable :: comment_line
    character(len=20) :: energy_line
    character(len=20) :: gnorm_line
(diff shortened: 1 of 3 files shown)


Albkat (approved):
LGTM.

PR #1350, opened 2025-11-07
Reduce OpenMP stack usage for GFN2 optimizations

foxtran:
Now, GFN1 and GFN2 can be run with `OMP_STACKSIZE=36kb` for systems with 300+ atoms.

Tested on input from #1338.
Tested commands:
```
OMP_STACKSIZE=36k OMP_NUM_THREADS=4 OMP_PLACES=24-27 ../xtb -P 4 --gfn 2 test.xyz  --opt --alpb water
OMP_STACKSIZE=36k OMP_NUM_THREADS=4 OMP_PLACES=24-27 ../xtb -P 4 --gfn 1 test.xyz  --opt --alpb water
OMP_STACKSIZE=36k OMP_NUM_THREADS=4 OMP_PLACES=24-27 ../xtb -P 4 --gfn 2 test.xyz  --opt
```

PR #1360, opened 2025-11-23
Fix OpenMP thread memory leaks in aespot, coulomb, disp, and hamiltonian

DanielWicz:
This pull request addresses a memory leak introduced after commit #1350. When running with Intel MKL, each GFN2 call increases memory usage by roughly 300–400 MB. After about 100 iterations, XTB reaches ~50 GB of RAM usage, whereas the same system required only ~1 GB before #1350.

After this commit it seems it comes back to normal, without any performance penalty. 

The issue appears to stem from several variables that are no longer being deallocated.

### DIFF
--- a/src/aespot.F90
+++ b/src/aespot.F90
@@ -410,6 +410,8 @@ subroutine mmompop_cpu(nat,nao,aoat2,xyz,p,s,dpint,qpint,dipm,qp)
 !$ qp(:,:) = qp + qp_omp
    !$omp end critical (qp_crt)
 
+!$ deallocate(dipm_omp, qp_omp)
+
    !$omp end parallel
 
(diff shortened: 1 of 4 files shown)


foxtran (approved):
LGTM.

foxtran:
Gfortran 15.2, Intel ifx 2025.2.0, latest Intel ifort puts deallocations... So, that is strange that deallocations are necessary...

toxtran:
Hm.. Thank you! I will check.

BTW, please, fix DSO check: https://github.com/grimme-lab/xtb/pull/1360/checks?check_run_id=56137660416

thfroitzheim (approved):
Thanks for the fix!

Issue #1280, opened 2025-05-12
Docking: `--nfinal 0` flag is ignored when `--fast` is specified afterward


marvinfriede:
**Describe the bug**
The `--nfinal 0` flag is ignored when `--fast` is specified afterward.

**To Reproduce**
- `xtb dock mol.xyz mol.xyz --nfinal 0 --fast` starts a GFN2 optimization.
- `xtb dock mol.xyz mol.xyz --fast --nfinal 0` correctly skips the optimization.
- `xtb dock mol.xyz mol.xyz --nfinal 0` correctly skips the optimization.

**Expected behaviour**
Argument order should not have an effect.

**Additional context**
Tested with a modified 6.7.1 version (41059b4).

PR #1364, opened 2025-11-24
fix issue 1280

chselz:
Should fix issue #1280 

I compiled it using meson and ifort compiler. For me it worked then using water as an example
```
3

O 0.000000 0.000000  0.000000
H 0.758602 0.000000  0.504284
H 0.260455 0.000000 -0.872893

```


lmseidler:
In `set_logicals` you could check using `getValue` whether the `--nfinal` flag was set, as you do for setting `nopt` in the first place in `set_docking` (where you now set the new logical). This avoids introducing the new logical.

PR #1368, opened 2025-12-04
extend --uhf keyword

### LINKED ISSUE #1162: ONIOM calc not passing # of unpaired electrons
I have been trying to set up and run ONION geometry-optimization calcs, and have been running into some problems.

If I run an open-shell system through xTB, it ignores my attempts to specify the number of unpaired electrons. I tried it using the input file ($spin 3) and I tried it in the command line (--uhf 3), and both times in the output file it gave

...................................................
: SETUP :
:.................................................:

      :  unpaired electrons                  0          :
      ...................................................

[CoTpOPhCl2.zip](https://github.com/user-attachments/files/18581214/CoTpOPhCl2.zip)

What am I doing wrong? How do I get it to run with the correct number of unpaired electrons?

chselz:
This should fix #1162 

If this PR is accepted then I will change docs

### DIFF
--- a/src/setparam.f90
+++ b/src/setparam.f90
@@ -107,7 +107,7 @@ module xtb_setparam
       !> inner region charge
       integer  :: innerchrg
 
-      !> explicite charges (inner:outer)
+      !> explicit charges (inner:outer)
       logical :: fixed_chrgs= .false.
       
       !> cut high order(>1) covalent bonds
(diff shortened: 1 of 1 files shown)


src/setparam.f90
lmseidler, line 1: explicite -> explicit

lmseidler (changes requested):
looks good but typo: explicite -> explicit

thfroitzheim (approved):

Issue #889, opened 2023-10-27
Misreported cycle number in `gradients` file


Andrew-S-Rosen:
**Describe the bug**
The `gradient` file generated with `--grad` always reports that the results are for cycle 1 even when they are (presumably) for a different cycle.

The top of the `gradient` file in the example below is as follows:

```
$grad
  cycle =      1    SCF energy =   -26.43824279253   |dE/dxyz| =  0.000389
```

Note how it says cycle 1. The SCF energy, based on the log report below, seems to be for the final calculation in the set of cycles and not the first.

**To Reproduce**
`xtb coord.xyz --opt --grad > output.log`

Output files below:

[samples.zip](https://github.com/grimme-lab/xtb/files/13193735/samples.zip)

**Expected behaviour**
The correct cycle number should be reported.

**Additional context**
None.

awvwgk:
The cycle reported here is a dummy rather than the actual cycle number.

Andrew-S-Rosen:
Yup. I just thought it could be potentially confusing to the user.

PR #1369, opened 2025-12-08
fix issue 889

chselz:
This should address issue #889 .

I made the iter_needed variable which tracks the number of needed iterations needed to complete the geometry optimization an optional variable, as this feature is only of interest for the --grad option and not for example the docking. Making it not optional would further neccesiate to change every geo opt driver call which is in my option not necessary. It is initialized to 0 and not 1 as also the geometry optimization iter starts counting at 0 and when for the unoptimized geometry the grad is requested this corresponds to iter = 0.

For me the build and testsuite worked using the meson build system with ifort compiler.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1014,7 +1014,7 @@ contains
 
       call generic_header(iprop, 'Property Printout', 49, 10)
       if (lgrad) then
-         call writeResultsTurbomole(mol, energy=etot, gradient=g, sigma=sigma, cycle=iter_needed)
+         call writeResultsTurbomole(mol, energy=etot, gradient=g, sigma=sigma, iter_number=iter_needed)
          if (allocated(basename)) then
             cdum = basename//'.engrad'
          else


src/io/writer/turbomole.f90
```diff
@@ -100,17 +101,18 @@ subroutine writeEnergyTurbomole(unit, energy)
 end subroutine writeEnergyTurbomole
 
 
-subroutine writeGradientTurbomole(unit, xyz, sym, energy, gradient)
+subroutine writeGradientTurbomole(unit, xyz, sym, energy, gradient, cycle)
```
foxtran: Why not `iter_number`? It is more clear than cycle.

chselz: As th TM writer anyway prints cycle = `cycle` I thought to name the variable `cycle` to stay consistent with the string that is printed out.

thfroitzheim: I agree with @foxtran that `cycle` is really an unfortunate naming, as it is an internal Fortran statement. Let's switch to something clearer

src/prog/main.F90
```diff
@@ -1013,7 +1014,7 @@ subroutine xtbMain(env, argParser)
 
       call generic_header(iprop, 'Property Printout', 49, 10)
       if (lgrad) then
-         call writeResultsTurbomole(mol, energy=etot, gradient=g, sigma=sigma)
+         call writeResultsTurbomole(mol, energy=etot, gradient=g, sigma=sigma, cycle=iter_needed)
```
toxtran: Here naming is wrong.

chselz: A thanks missed that one

toxtran:
@chselz, did you check code with non-default optimizers, for example...

Issue #1286, opened 2025-05-16
`bench.xyz/1707_0569xWATER.xyz` can not  be optimized with GFN-FF


foxtran:
**Describe the bug**
Running of GFN-FF optimization of bench.xyz/1707_0569xWATER.xyz fails with:
```
 * calculating model hessian...
Using GFN-FF Lindh-Hessian
 * fragmented diagonalization... 2 fragments
 Shifting diagonal of input Hessian by    2.0062370709201787E-002
 Lowest  eigenvalues of input Hessian
   -0.000062   -0.000024    0.000000    0.000000    0.000000    0.000000
    0.000000    0.000000    0.000000    0.000000    0.000000    0.000000
    0.000000    0.000000    0.000000    0.000000    0.000000    0.000000
 Highest eigenvalues
    0.455528    0.455639    1.353426    1.353680    1.461000    1.461173

 k=          18   nvar=        5121
 ANC generation failed
```

At the same time, geometry optimization with GFN2 starts:
```
 * calculating model hessian...
Using Lindh-Hessian (1995)
 Shifting diagonal of input Hessian by    1.6758195813745262E-002
 Lowest  eigenvalues of input Hessian
    0.010000    0.011217    0.011716    0.011716    0.012114    0.012630
    0.013063    0.013087    0.013463    0.013491    0.013663    0.014041
    0.014349    0.014389    0.014667    0.014703    0.014828    0.015191
 Highest eigenvalues
    1.360496    1.360685    1.363306    1.363780    1.363947    1.379580
```

**To Reproduce**
Steps to reproduce the behaviour:

1. See #1258 
2. GFN-FF: `../xtb --gfnff bench.xyz/1707_0569xWATER.xyz --opt --cycles 30 -P 1`
3. GFN-2: `../xtb --gfn 2 bench.xyz/1707_0569xWATER.xyz --opt --cycles 30 -P 1`

**Expected behaviour**

Start geometry optimization with GFN-FF

Issue #800, opened 2023-04-05
Failing Geometry optimization with fixed atoms using the GFN-FF


benbaed:
**Describe the bug**
Geometry optimizations using the GFN-FF do not work when I try to fix certain atoms of the structure. 
A possible work-around is to use the ANC optimizer and not the default L-ANC optimizer. 

**To Reproduce**

This consistently happens with multiple of my NCI structures (I could not reproduce it using Benzene)
1. start xtb (version 6.5.1) with:
 ``` xtb struc.xyz --gfnff --gbsa toluene --opt vtight --input xtb.inp > xtb.out ```

2. the following errors are printed:
[ERROR] Program stopped due to fatal error
-3- Geometry optimization failed
-2- xtb_geoopt: Geometry optimization did not converge
-1- gfnff_setup: Could not read topology file.

3.  Searching for "L-ANC optimizer", one can find the following message:
k=         528   nvar=        4974
 ANC generation failed

**Expected behaviour**
As using the normal ANC optimizer, the given atoms should be fixed and the optimization should be running and not stop.

**Additional context**
I found that using the normal ANC optimizer solves this issue, the xtb.inp then looks like this:
```
$opt
 engine=rf
$fix
  atoms: 1-1482
$end 
```
Since the ANC optimizer is the default for gfn2, I did not observe the bug there.

**Input and generated files**
[gfnff_fix_bug.zip](https://github.com/grimme-lab/xtb/files/11158982/gfnff_fix_bug.zip)

marcelmbn:
@benbaed @Thomas3R Is this issue still present?

PR #1379, opened 2026-01-20
Failure amid $fix atoms and many disconnected fragments when optimizing with the L-ANC

DanielWicz:
I encountered an optimization problem when using $fix. I found that it is related to bugs #1286  and #800 . I attempted to address the issue, and my solution appears to work. It seems that when there are many fragments and/or multiple $fix statements, zero eigenvalues appear, which then cause problems downstream in the optimization protocol. I tried to numerically “patch” this issue, and it seems to work, but it requires additional babysitting and checks to ensure that no artifacts are introduced.

I also added a unit test. The test fails if the patch introduced by the commit is not present. However, the unit test itself also requires careful review.


cmake/modules/xtb-utils.cmake
```diff
@@ -25,6 +25,15 @@ macro(
...
 
+# If the dependency was already added by another subproject (e.g., via
+# FetchContent) it might only provide the plain target `${package}` without the
+# namespaced alias `${package}::${package}` expected by xTB. Create the alias
+# early to avoid re-adding the project and triggering duplicate-target errors.
+if(TARGET "${package}" AND NOT TARGET "${package}::${package}")
+   add_library("${package}::${package}" INTERFACE IMPORTED)
+   target_link_libraries("${package}::${package}" INTERFACE "${package}")
+endif()
+
```
foxtran, line 36: Please, consider to create  a separate PR for this.

PR #1381, opened 2026-01-27
Implementation of O1NumHess algorithm and model Hessian refactor

lmseidler:
This pull request introduces support for a new Hessian calculation method using the ODLR (O1NumHess) approximation. The main changes include adding the new `--o1nh` command-line option, implementing the ODLR Hessian algorithm, and updating the workflow and output to accommodate this new method. The implementation ensures that the ODLR Hessian is only used when appropriate and disables IR/Raman intensity calculations when this method is selected.

**Key changes:**

### ODLR Hessian Implementation
- Added a new `odlrhessian` subroutine to the `TCalculator` class, implementing the ODLR (O1NumHess) Hessian calculation based on Wang et al. (https://doi.org/10.1021/acs.jctc.5c01354).
- Integrated the new ODLR Hessian logic into the main Hessian workflow in `numhess`, calling `odlrhessian` when the new method is selected.
- Utility functions implemented in new o1numhess module. This includes an implementation of a modified Swart model Hessian (I couldn't get the one used in ANCOPT to work) and a generic iterative CG solver, etc.
- Added Ref. to the startup description.

### User Interface and Workflow
- Introduced the `--o1nh` command-line option to enable the ODLR Hessian calculation, and documented it in the manual. [[1]](diffhunk://#diff-9ca77c981033cb0bbe604a1d423b77f1b54b50b2ff870cafa951e33fe34e084eR237-R239) [[2]](diffhunk://#diff-6814998c6fca247a3979151ab1b954221ab76554d2a8bcd22f8f6c03c58612c4R1871-R1873)
- Added the `o1numhess` logical flag to the settings structure, and ensured it is properly parsed and propagated throughout the codebase.
- Updated the main program logic to check for compatibility (e.g., no frozen atoms or PTB) when ODLR Hessian is requested, and to issue appropriate warnings or errors.

### Output and Feature Adjustments
- ~~Disabled IR and Raman intensity calculations and related output when the ODLR Hessian method is used, as these are not supported by this approach.~~ [[1]](diffhunk://#diff-565fb80b708ce0e4f18fcaee3a834915ceb64970c6e0911b3c0f9b59829862d5R506-R507) [[2]](diffhunk://#diff-4592fbb9a1e59ee7b1c9b631739366a9b9c1434b63a013a5844a21142280b780R655-L659)
- Added informative output and error messages for unsupported combinations and recommendations for step size when using ODLR Hessian.

### Codebase and Build System
- Added the new source file `o1numhess.f90` to the build system.
- Refactored and extended module imports and procedure declarations to support the new functionality. [[1]](diffhunk://#diff-7b72c6d4d9e02e7c02ed7bee9475629eb9f64f26f2df31fcb906931fff591119R20-L31) [[2]](diffhunk://#diff-7b72c6d4d9e02e7c02ed7bee9475629eb9f64f26f2df31fcb906931fff591119R52-R54)

### Miscellaneous
- Small improvement in error handling for edge cases (e.g., frozen atoms) and added TODO comments for future improvements. [[1]](diffhunk://#diff-565fb80b708ce0e4f18fcaee3a834915ceb64970c6e0911b3c0f9b59829862d5L464-R469) [[2]](diffhunk://#diff-565fb80b708ce0e4f18fcaee3a834915ceb64970c6e0911b3c0f9b59829862d5R142) [[3]](diffhunk://#diff-6814998c6fca247a3979151ab1b954221ab76554d2a8bcd22f8f6c03c58612c4R549)
- Included basic fixtures system for the hessian test suite (this can be easily reused for other cases in the future).


foxtran:
This PR is really hard to check. Please, rebase all your changes to the current main branch.
It should look like https://github.com/grimme-lab/xtb/compare/main...foxtran:fixtran/1381. I spent about 2 minutes to do so including cloning of repos.

You can easily do it with the following sequence of commands:
```console
git clone git@github.com:grimme-lab/xtb.git xtb.1381
cd xtb.1381
git remote add lmseidler git@github.com:lmseidler/xtb.git
git fetch lmseidler
git checkout lmseidler/o1numhess
git rebase origin/main
# I use kdiff3 for resolving conflicts
git mergetool --tool=kdiff3
# solve conflicts in src/extern/orca.f90; take 3rd variant of code with inquire.
git rebase --continue # continue after stopping
# I created  a new branch fixtran/1381  below; you should check it before overwriting lmseidler/o1numhess. 
# Hope, you will not need in `git reflog`
git checkout -b fixtran/1381 
git push lmseidler
```

assets/references.bib
```diff
@@ -203,3 +203,14 @@ @article{grimme_non-self-consistent_2023
 	pages = {124111},
 	file = {Full Text PDF:/Users/marcelmueller/Zotero/storage/LRJJ6T9I/Grimme et al. - 2023 - A non-self-consistent tight-binding electronic str.pdf:application/pdf;Snapshot:/Users/marcelmueller/Zotero/storage/FAT76IUH/A-non-self-consistent-tight-binding-electronic.html:text/html},
 }
+
+@article{wang2025o1numhess,
+  title={O1NumHess: a fast and accurate seminumerical Hessian algorithm using only O (1) gradients},
+  author={Wang, Bo and Luo, Shaohang and Wang, Zikuan and Liu, Wenjian},
+  journal={Journal of Chemical Theory and Computation},
```
chselz: The other papers have their journal abbreviated, i think we should stay consistent in this regard

src/prog/main.F90
```diff
@@ -209,6 +209,7 @@ subroutine xtbMain(env, argParser)
       logical :: strict
       logical :: exitRun
       logical :: cold_fusion
+      logical :: odlr_valid
```
chselz: i think its not a good idea to put this variable into the undocumented and unexplainable variable block of variables. maybe move it to a better place in the declaration section and add a comment what this variable holds / what it is used for

src/CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ list(APPEND srcs
   "${dir}/modef.f90"
   "${dir}/model_hessian.f90"
   "${dir}/neighbor.f90"
+  "${dir}/o1numhess.f90"
```
chselz, line 119: in the long run we should think about putting all the methods to calculate hessians e.g. model_hessian.f90, o1numhess.f90 etc into one seperate directory, as now the dir where these files are currently located in gets even more populated and therefore less understandable

src/setparam.f90
```diff
@@ -343,6 +343,8 @@ module xtb_setparam
    real(wp) :: step_hess = 0.005_wp
    ! Scaling factor for the hessian elements
    real(wp) :: scale_hess = 1.0_wp
+   ! Use O1NumHess for Hessian
+   logical :: o1numhess = .false.
```
chselz, line 347: probably also in the long run we should change for the hessians to something similar as its done with the method selection that we dont have many logicals but one integer variable that holds an integer that assigns the hessian method

src/hessian.F90
```diff
@@ -111,7 +111,7 @@ subroutine numhess( &
    real(wp),allocatable :: pold(:)
    real(wp),allocatable :: dipd(:,:), dalphadr(:,:), dalphadq(:,:)
    real(wp),allocatable :: amass_au(:), amass_amu(:)
-   real(wp) :: asq, gamsq
+   real(wp) :: asq, gamsq, final_err
```
chselz: leave a comment describing your variable

src/type/calculator.f90
```diff
@@ -227,4 +231,310 @@ subroutine hessian_point(self, env, mol0, chk0, iat, ic, step, energy, gradient,
...
+   real(wp), intent(in) :: step
+   !> Array to add Hessian to
+   real(wp), intent(inout) :: hess(:, :)
+   !> Array for displacement directions
+   real(wp), allocatable :: displdir(:, :)
+   !> Final error after all steps
+   real(wp), intent(out) :: final_err
+   
+   ! UFF vdw radii - could be replaced with any other vdw radii i guess
+   real(wp), parameter :: vdw_radii(103) = [ &
```
chselz: i am not too much of a fan having the parameters here, i would use radii implemented in mctc lib or put somewhere into the parameters implemented in this code base

test/unit/fixtures/nblist_ref
chselz, line 1: i dont understand this file, what it does, what its contents mean and why its basically the same numbers repeated over and over again

test/unit/fixtures/distmat_ref
chselz, line 1: i dont understand this either

src/o1numhess.f90
```diff
@@ -0,0 +1,957 @@
...
+      real(wp), allocatable :: tmp(:, :)
+      real(wp), allocatable :: tmp2(:, :)
+      real(wp), allocatable :: f1(:, :)
+      real(wp), allocatable :: f2(:, :)
+   end type odlr_operator_data
+
+   character(len=*), parameter :: source = 'xtb_o1numhess'
+
+   !> covalent radii
+   real(wp), parameter :: cov(103) = [&
```
chselz: these should also be used from mctc-lib

src/o1numhess.f90
chselz, line 1: please add docstrings for each subroutine/function

chselz (changes requested):
Some minor things i would correct
however i did not look into implementation details of o1numhess itself, i.e. checking if theory itself was implemented correctly

test/unit/fixtures/nblist_ref
foxtran: This file contains reference answers. Nothing special.

src/CMakeLists.txt
lmseidler: yes but i think this is part of the future heap of work

src/setparam.f90
lmseidler: agreed but again future work

test/unit/fixtures/distmat_ref
lmseidler: again test fixtures

src/type/calculator.f90
lmseidler: moved them into separate file

src/o1numhess.f90
lmseidler: i kept them because the ones in mctc-lib are modified for smaller radii for metals. these should be the original ones for the most part

chselz (dismissed):
I think it should be fine now

Before merging probably someone else should also approve as this is quite a big PR

src/type/calculator.f90
```diff
@@ -301,7 +293,13 @@ subroutine odlrhessian(self, env, mol0, chk0, step, hess, final_err)
    do i = 1, mol0%n
       do j = i, mol0%n
          ! effective distmat
-         dist = mol0%dist(i, j) - vdw_radii(mol0%at(i)) - vdw_radii(mol0%at(j))
+         ri = get_cov_rad(mol0%at(i))
+         rj = get_cov_rad(mol0%at(j))
+         if (ri < 0.0_wp .or. rj < 0.0_wp) then
```
toxtran, line 341: This check can be done once at start of program

src/o1numhess.f90
```diff
@@ -1,6 +1,6 @@
 ! This file is part of xtb.
 !
-! Copyright (C) 2017-2020 Stefan Grimme
```
toxtran: What? That is not your code.

Add extra line instead of replacing

lmseidler: then i don't understand how there are other names and years in other files. since i wrote all the code in this file, i should also have the copyright for the code under gpl.

src/type/calculator.f90
lmseidler: i would still prefer to have the check at this point because otherwise you might get an unhandled segfault with no info about where or why it happened.

edit: the computational cost should also be very low so this is mostly a question of error handling imo

src/o1numhess.f90
foxtran: Sorry... It is a new file. I checked a commit, not the full diff. Then, it  is OK.

src/param/covalentrad.f90
```diff
@@ -0,0 +1,56 @@
...
+      & 1.36_wp, 1.34_wp, 1.22_wp, 1.19_wp, 1.16_wp, 1.10_wp, 1.11_wp, 1.12_wp, 1.18_wp, 1.24_wp, 1.21_wp, &
+      & 1.21_wp, 1.16_wp, 1.14_wp, 1.17_wp, 2.10_wp, 1.85_wp, 1.63_wp, 1.54_wp, 1.47_wp, 1.38_wp, 1.28_wp, &
+      & 1.25_wp, 1.25_wp, 1.20_wp, 1.28_wp, 1.36_wp, 1.42_wp, 1.40_wp, 1.40_wp, 1.36_wp, 1.33_wp, 1.31_wp, &
+      & 2.32_wp, 1.96_wp, 1.80_wp, 1.63_wp, 1.76_wp, 1.74_wp, 1.73_wp, 1.72_wp, 1.68_wp, 1.69_wp, 1.68_wp, &
+      & 1.67_wp, 1.66_wp, 1.65_wp, 1.64_wp, 1.70_wp, 1.62_wp, 1.52_wp, 1.46_wp, 1.37_wp, 1.31_wp, 1.29_wp, &
+      & 1.22_wp, 1.23_wp, 1.24_wp, 1.33_wp, 1.44_wp, 1.44_wp, 1.51_wp, 1.45_wp, 1.47_wp, 1.42_wp, 2.23_wp, &
+      & 2.01_wp, 1.86_wp, 1.75_wp, 1.69_wp, 1.70_wp, 1.71_wp, 1.72_wp, 1.66_wp, 1.66_wp, 1.68_wp, 1.68_wp, &
+      & 1.65_wp, 1.67_wp, 1.73_wp, 1.76_wp, 1.61_wp] / autoaa
+
+   ! NOTE: these should later be moved to mctc-lib
```
thfroitzheim, line 41: Can't we just use the mctc-lib radii directly and scale them back to their original value?

src/param/uffvdwrad.f90
```diff
@@ -0,0 +1,54 @@
...
+! (at your option) any later version.
+!
+! xtb is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
+!> UFF vdw radii - could be replaced with any other vdw radii i guess
```
thfroitzheim, line 17: Should we put those into mctc-lib or just use our vdw radii?

src/prog/main.F90
```diff
@@ -241,7 +243,17 @@ subroutine xtbMain(env, argParser)
...
-         if(set%mode_extrun == p_ext_ptb) then
+         ! O1NumHess only supports Hessian calculation, nothing else
+         odlr_valid = freezeset%n == 0 .and. set%mode_extrun /= p_ext_ptb
+         if (set%o1numhess) then
+            if (.not. odlr_valid) then
+               call env%error("O1NumHess does not support frozen atoms/PTB", source)
+            end if
+            call env%warning("O1NumHess does not support calculation of IR or Raman activities", source)
+            if (set%step_hess >= 0.0005_wp) then
+               call env%warning("Step size for O1NumHess can be chosen to be much smaller (recommended: 1.0e-6)", source)
```
thfroitzheim: 1e-6 seems very small. Is our SCF convergence even tight enough for this?

src/type/calculator.f90
```diff
@@ -227,4 +233,306 @@ subroutine hessian_point(self, env, mol0, chk0, iat, ic, step, energy, gradient,
...
+   !> Step size for numerical differentiation
+   real(wp), intent(in) :: step
+   !> Array to add Hessian to
+   real(wp), intent(inout) :: hess(:, :)
+   !> Array for displacement directions
+   real(wp), allocatable :: displdir(:, :)
+   !> Final error after all steps
+   real(wp), intent(out) :: final_err
+   
+   real(wp), parameter :: dmax = 1.0_wp, eps = 1.0e-8_wp, eps2 = 1.0e-15_wp, imagthr = 1.0e-8_wp
```
thfroitzheim: should those really be hard-coded constants? Can we make them dependent on the epsilon for a floating point operation?

src/type/calculator.f90
```diff
@@ -227,4 +233,306 @@ subroutine hessian_point(self, env, mol0, chk0, iat, ic, step, energy, gradient,
...
+      ndispl0 = ndispl_final
+      ndispl_final = ndispl_final + nimg
+      call get_gradient_derivs(self, env, step, ndispl0, ndispl_final, displdir, mol0, chk0, g0, .false., g)
+      call gen_local_hessian(env, ndispl_final, distmat, displdir, g, dmax, hess)
+      call lr_loop(env, ndispl_final, g, hess, displdir, final_err)
+   end if
+
+end subroutine odlrhessian
+
+subroutine get_gradient_derivs(self, env, step, ndispl0, ndispl_final, displdir, mol0, chk0, g0, doublesided, g)
```
thfroitzheim: Can we use this consistently also in the normal hessian?

src/o1numhess.f90
```diff
@@ -0,0 +1,981 @@
...
+   do i = 1, nat
+      do j = i + 1, nat
+         Hint = 0.35_wp * screenfunc(i, j)**3
+         bmat6 = bmat_bond(xyz(:, i) - xyz(:, j))
+         outer6 = (spread(bmat6, dim=2, ncopies=6) * spread(bmat6, dim=1, ncopies=6))
+         i1 = 3 * i - 2
+         i2 = 3 * i
+         j1 = 3 * j - 2
+         j2 = 3 * j
+         hess_local(i1:i2, i1:i2) = hess_local(i1:i2, i1:i2) + Hint * outer6(1:3, 1:3)
```
thfroitzheim: Please add some explanation on what is happening here

test/unit/test_hessian.f90
```diff
@@ -36,14 +40,41 @@ module test_hessian
...
 contains
 
+function get_fixture_path(filename) result(fullpath)
+   character(len=*), intent(in) :: filename
+   character(len=:), allocatable :: fullpath
+   character(len=1024) :: base_dir
+   integer :: length, stat
+   
+   ! Get variable; status 0 means it exists
+   call get_environment_variable("TEST_FIXTURES_DIR", base_dir, length, stat)
```
thfroitzheim: This is just for the reference neighbor list and distance matrix? Can't we add to the test directory a data folder where they are always stored. Then we don't need to do any path resolution

thfroitzheim (changes requested):

DanielWicz:
@lmseidler Hey, I have a working implementation of O1HessFast since few months for xtb.

I didn't have time to open pull request. But maybe you could benefit from using my implementation make it work or copy/paste part of it? The core difference is that there that this version is heavily optimized for the parallel calculations, to make it extremely fast. 


The code is under https://github.com/DanielWicz/xtb/tree/calcnested 

It starts with commit: https://github.com/DanielWicz/xtb/commit/0a8b51c81c2b4c75e56a33ee81d15e2bb5341d03
And most of the corrections are until commit: https://github.com/DanielWicz/xtb/commit/1f06a3a84d08ac6234dad7ab9cee10ac99cd44f0

There are few additional steps added (additional sampling of degrees of freedom) to remove imaginary frequencies on top of the method. 

Upon compilation it works like this: xtb structure.xyz --ohess (or --hess) --onehessfast


Or if anyone has some ideas, I can e.g. try to merge with @lmseidler. I just need to know how functional is your current code.

lmseidler:
Hi, the current version is basically usable but not very polished, which I didn't get to yet. What would definitely interesting if you managed to get an efficiently parallelized version to get the displacement vectors? This is currently kind of a bottleneck in the parallelization. 

I will check out your version and compare, also for other things, then maybe we can merge some stuff.

lmseidler:
Hi, first of all sorry to everyone who has to review this. There are a lot of changes so I'll try to summarize it here:

## 1. O1NumHess Implementation
This is obviously the main part. Changes are localized in the tests, and in `src/o1numhess.f90`, `src/hessian.F90`, as well as `src/type/calculator.f90`. This also modified the existing Hessian implementation because both algorithms now share a common gradient evaluator. The unit tests for the Hessians are still passing, so functionally everything stayed the same. 

With the current implementation, I managed to improve wall times significantly compared to my first version:

Taxol with `OMP_NUM_THREADS=1`: 87.2s -> 49.6s
" `OMP_NUM_THREADS=28`: 26.9s -> 7.9s

This is now matching or faster than your implementation @DanielWicz , however for larger molecules I would still expect your implementation to be slightly faster.

## 2. Model Hessian Refactor
This is the second part and is probably responsible for the most changed lines. Previously, the model Hessians were implemented in a very messy way in `src/model_hessian.f90`. These implementations were partially duplicated and also contained a handful of bugs, which are now fixed. I replaced this old version by making a more modular, decoupled implementation in `src/model_hessian/`, fitting common patterns in our newer projects, e.g. tblite. This significantly cuts down on code lines and makes everything more readable/streamlined. I think all callers were updated to use the new interfaces and signatures. In the refactoring process I used tight unit tests to make sure that the values stay consistent. The fixtures probably bloat the number of lines committed in this PR significantly. 

The only significant changes can be found in the Swart model Hessian, because this originally contained two bugs that lead to significant deviations. I'm not sure if these affected ANCOPT runs. My test calculations ran without problems, however. Finally, there are some weird things that I found here, which I commented on using `! NOTE:`. Probably someone should take a look at these.

As a side note, `src/bmatrix.f90` contains helpers for analytical Wilson B-matrix setups (internal coordinate derivatives). This will later be used for the implementation of compliance elements.

## Things left TODO
- [x] Hessian C-API is broken right now. This is also true for the main branch. I will fix this here.
- [x] I need to run more test calculations to make sure there are no weird segfaults or obvious errors.
- [x] Obviously something still needs to be done about the CI runs.

lmseidler:
While running some more tests I fixed several other things:
- stack overflow from symmetry symbols
- added some better handling for edge cases in o1numhess
- small latent Hessian issues

The change in broyden.f90 is maybe a bit questionable. You should probably check that quick @thfroitzheim . Otherwise, I would consider this done and ready to merge for now.

So far results look good: GMTKN55 and S30L now run without issues. Median total CPU speedup for S30L is 20x (range: roughly 13x-39x).

lmseidler:
I verified geom opts and docking. Geometry optimizations of multiple large molecules from the test suite generally have a final RSMD < 5e-3 A, energies and average number of iterations don't really change. Docking for multiple different cases always leads to different geometries, but energies agree < 0.1 kJ/mol. The reason for the different geometries here is mostly the chaotic nature of the genetic search in the docking algorithm. Some small difference in the model Hessians is amplified hugely in that step. Overall, this should sufficiently validate the changes.

src/type/calculator.f90
```diff
@@ -134,97 +172,544 @@ subroutine hessian(self, env, mol0, chk0, list, step, hess, dipgrad, polgrad)
...
+
+   call mol%copy(mol0)
+   call chk%copy(chk0)
+
+   ! hessian initial guess
+   allocate (h0(N, N))
+   block
+      type(TSwartModelHessian) :: model_hessian
+      type(modhess_setvar) :: modh
+      modh = modhess_setvar(kr=0.4000_wp, kf=0.1300_wp, kt=0.0075_wp, &
```
thfroitzheim, line 315: Please add a comment on where those parameters come from

src/broyden.f90
```diff
@@ -108,7 +108,16 @@ subroutine broyden(n,q,qlast,dq,dqlast,iter,maxiter,&
 
    !     Build dF(iter-1)
    df(it1,:)=dq-dqlast
-   inv=1.0d0 / sqrt(dot_product(df(it1,:),df(it1,:)))
+   inv=sqrt(dot_product(df(it1,:),df(it1,:)))
```
thfroitzheim, line 111: Not sure this is the right logic. If the residual is almost 0, the SCF is converged. In this case the weight of this iteration goes to infinity (it is just the right state). This is not a case where we want to do a simple damping step and go only e.g. 40% in this direction. If this is actually ever a problem, I would either regularize the inverse with a small number or take a full step to the new charges (so alpha=1)

src/hessian.F90
```diff
@@ -185,7 +190,21 @@ subroutine numhess( &
...
+      if (set%elprop == p_elprop_alpha) then
+         call calc%hessian(env, mol, chk0, indx, step, h, dipd, dalphadr, &
+            & odlr=.true., final_err=final_err)
+      else
+         call calc%hessian(env, mol, chk0, indx, step, h, dipd, odlr=.true., final_err=final_err)
+      end if
+      call env%check(exitRun)
+      if (exitRun) return
+      write(env%unit, '(A,1X,ES12.4)') "Error norm for predicted gradient (ODLR Hessian):", final_err
+   else if(freezeset%n.gt.0) then
```
thfroitzheim, line 207: What happens for freeze + O1NumHess? Is this case correctly handled in the input?

src/meson.build
```diff
@@ -106,8 +106,15 @@ srcs += files(
   'mdoptim.f90',
   'metadynamic.f90',
   'modef.f90',
-  'model_hessian.f90',
```
thfroitzheim, line 109: Let's add `model_hessian` as a folder with its own meson build file

src/o1numhess.f90
```diff
@@ -0,0 +1,779 @@
...
+   !> Displacement directions
+   real(wp), intent(in), target :: displdir(:, :)
+   !> Gradient derivatives
+   real(wp), intent(in) :: g(:, :)
+   !> Maximum distance threshold
+   real(wp), intent(in) :: dmax
+   !> Output Hessian matrix
+   real(wp), intent(out) :: hess_out(:, :)
+
+   real(wp), parameter :: lam = 1.0e-2_wp, bet = 1.5_wp, ddmax = 5.0_wp
```
thfroitzheim, line 87: What do those numbers mean? Add a comment

src/o1numhess.f90
```diff
@@ -0,0 +1,779 @@
...
+   !> Number of neighbors per atom
+   integer, intent(in) :: nbcounts(n)
+   !> Thresholds for sign determination and convergence
+   real(wp), intent(in) :: eps, eps2
+   !> Displacement directions
+   real(wp), intent(inout) :: displdir(n, n)
+   !> Final number of displacements
+   integer, intent(out) :: ndispl_final
+   !> LAPACK error code
+   integer, intent(out) :: ierr
```
thfroitzheim, line 443: Can we translate the error code directly into an actual error?

src/o1numhess.f90
```diff
@@ -0,0 +1,779 @@
...
+
+   deallocate(eye, locev_store)
+end subroutine gen_displdir
+
+!> Find significant imaginary modes of a Cartesian Hessian using the same
+!> projection (trproj, unit-mass geometric center) and mass weighting as the
+!> vibrational frequency path in src/hessian.F90. Returns Cartesian
+!> displacement directions and frequencies in cm^-1 (negative = imaginary).
+!> Linear molecules skip projection (matching src/hessian.F90)
+subroutine find_projected_imag_modes(env, mol, hess, linear, max_modes, modes, freqs, nmodes)
```
thfroitzheim, line 690: Is this handling of the imaginary modes consistent with the way we handle rotations and vibrations in the normal hessian? Is this actually O1NumHess specific or should we select one approach in general?

test/unit/fixtures/model_hessian/gff_h2o.dat
thfroitzheim, line 1: Is there an actual benefit to storing those fixtures outside of the actual test? You could also add a formatted array with three columns into the test, and one would directly know what those numbers mean (plus it would be only 1/3 the size)

test/unit/CMakeLists.txt
```diff
@@ -92,15 +94,15 @@ foreach(t IN LISTS tests)
   set_tests_properties(
     "${PROJECT_NAME}/${t}"
     PROPERTIES
-    ENVIRONMENT XTBPATH=${PROJECT_SOURCE_DIR}
+    ENVIRONMENT "XTBPATH=${PROJECT_SOURCE_DIR};TEST_FIXTURES_DIR=${CMAKE_CURRENT_SOURCE_DIR}/fixtures"
```
thfroitzheim, line 97: As written above, I don't really see the benefit in unformatted files holding test results

test/unit/main.f90
```diff
@@ -57,8 +59,9 @@ program tester
    stat = 0
 
    testsuites = [ &
-      new_testsuite("atomlist", collect_atomlist), &
-      new_testsuite("coordinationnumber", collect_coordinationnumber), &
+       new_testsuite("atomlist", collect_atomlist), &
+       new_testsuite("bmatrix", collect_bmatrix), &
+       new_testsuite("coordinationnumber", collect_coordinationnumber), &
```
thfroitzheim, line 64:
```suggestion
      new_testsuite("atomlist", collect_atomlist), &
      new_testsuite("bmatrix", collect_bmatrix), &
      new_testsuite("coordinationnumber", collect_coordinationnumber), &
```

test/unit/meson.build
```diff
@@ -32,8 +32,9 @@ if not testdrive_dep.found()
 endif
 
 tests = [
```
thfroitzheim, line 34: Fix the indentation also here

thfroitzheim (changes requested):

src/o1numhess.f90
lmseidler: mostly yes, i just saw that the projection in numhess uses some global masses, i'm not sure if we should use these here as well. i see that in other cases these are also not used but i guess these are only supposed to be used for the mass-weighted hessian?

lmseidler:
i added two new cml args to configure the imaginary frequency repair section of the implementation. otherwise, most issues should be addressed.

thfroitzheim (approved):

PR #1396, opened 2026-04-08
Fix macOS actions

marvinfriede:


foxtran:
That is memory corruption. Should be reproducible on Linux.

foxtran:
Ah no... That is something strange with OpenBLAS.

foxtran:
It is a bug in OpenBLAS. Fix is here: https://github.com/OpenMathLib/OpenBLAS/pull/5753

foxtran:
Fixed in upstream of OpenBLAS

marvinfriede:
Nice find. I guess we wait on the 0.3.33 release.

.github/workflows/fortran-build.yml
```diff
@@ -33,16 +33,16 @@ jobs:
          - name: Install OpenBLAS (macOS)
            if: ${{ contains(matrix.os, 'macos') }}
            run: |
-            brew install openblas
-            echo "PKG_CONFIG_PATH=/usr/local/opt/openblas/lib/pkgconfig" >> $GITHUB_ENV
-            echo "LDFLAGS=-L/opt/homebrew/opt/openblas/lib" >> $GITHUB_ENV
-            echo "CPPFLAGS=-I/opt/homebrew/opt/openblas/include" >> $GITHUB_ENV
+            brew install openblas@0.3.31
```
thfroitzheim:
```suggestion
            brew install openblas
```

thfroitzheim (changes requested):

.github/workflows/fortran-build.yml
foxtran: Yep, there is no pinning

foxtran:
@thfroitzheim, there is no pinning for OpenBLAS in brew. Contribution into brew are welcome :-)

thfroitzheim:
The default version on brew seems to have been bumped to 0.3.33 and it works again.

PR #1397, opened 2026-04-13
Activate solvation, electric fields, constraints and some properties with tblite

thfroitzheim:
Activates the now fixed solvation with tblite. To be consistent with tblite, I added the non-parametrized versions of GBSA and ALPB. The electric field should have been working all along, but I also activated it through the CLI.

For constraints the required additional potentials were simply missing. I also added a summary output, which contains the constraint energy. 

Since we had no properties at all with tblite, I added the basic stuff via the tblite post-processing (additional properties will follow in further PRs).


src/vertical.F90
```diff
@@ -14,6 +14,10 @@
 ! You should have received a copy of the GNU Lesser General Public License
 ! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
 
+#ifndef WITH_TBLITE
+#define WITH_TBLITE 0
+#endif
```
chselz, line 19: after seeing this statment now many times how about we make tblite not an option anymore thus we can get rid of all of these statments, as for gxtb tblite is mandatory

src/xhelp.f90
```diff
@@ -177,6 +181,14 @@ subroutine help(iunit)
    "    The solvent input is not case-sensitive.", &
    "    The Gsolv reference state can be chosen as reference, bar1M, or gsolv (default).",&
    "",&
+   "--gbe SOLVENT/EPSILON",&
+   "    Use generalized Born for finite epsilon (GBE) solvation model (tblite required).", &
+   "    Solvent is specified by dielectric constant or the solvent name.",&
+   "",&
```
chselz, line 187: when tblite with alpb is specified no non electrostatics are included right? maybe this should be added to the --alpb flag

src/tblite/calculator.F90
```diff
@@ -117,6 +140,17 @@ module xtb_tblite_calculator
    !> Conversion factor from temperature to energy
    real(wp), parameter :: kt = 3.166808578545117e-06_wp
```
chselz, line 141: with mctc constants here availible could we get rid of this hardcoded constant?

src/tblite/calculator.F90
```diff
@@ -201,29 +236,176 @@ subroutine newTBLiteCalculator(env, mol, calc, input)
...
+         kernel = merge(born_kernel%still, born_kernel%p16, alpb)
+
+         if (.not.parametrized_solvation .and. sol_state /= solution_state%gsolv) then
+            call fatal_error(error, "Solution state shift is only supported for named solvents")
+            return
+         end if
+
+         ! Construct purely electrostatic solvation model input for tblite
+         solv_input%alpb = alpb_input(solvent%eps, kernel=kernel, alpb=alpb)
+      case("cosmo")
```
chselz, line 396: cosmo keyword and then cpcm error message and comments is not consistent as these are in principle two different models (zeta scaling factor is different -right?). Further is this then cpcm or ddcpcm?

src/tblite/calculator.F90
```diff
@@ -511,9 +775,32 @@ subroutine num_grad_chrg(env, mol, tblite)
    call close_file(ich)
    write(env%unit, '(1x, a)') "CEH gradients written to ceh.charges.numgrad"
 
- end subroutine num_grad_chrg
+end subroutine num_grad_chrg
  
 
+subroutine get_argument_as_real(arg, val, error)
+   !> Index of command line argument, range [0:command_argument_count()]
```
chselz: why do we have this function in the calculator file? would it not make sense to use from mctc-lib and further why is it used here? doesn't the input parsing happens in prog/main.f90

src/prog/main.F90
```diff
@@ -1045,6 +1051,8 @@ subroutine xtbMain(env, argParser)
             call main_property(iprop, env, mol, chk%wfn, calc%basis, calc%xtbData, res, &
                & calc%solvation, set%acc)
             call main_cube(set%verbose, mol, chk%wfn, calc%basis, res)
+         type is (TTBLiteCalculator)
+            call tblite_property(iprop, env, chk, calc, mol, res)
```
chselz, line 1055: while i think this is a way to go here to have tblite have its own property printout, and probably way easier to implement for now 😅  would it in principle be also possible to parse the tblite result into the main_property routine thus we would also get rid of some code im the calculator file

chselz (dismissed):
LGTM, i think if we use tblite as the sole backend for hamiltonians a lot of stuff will clear up from what now seems a bit hacky.

src/vertical.F90
thfroitzheim: Yes, I think we should do this, but not right now. First, we should release a final version in the current form

src/xhelp.f90
thfroitzheim: No, with `--alpb` tblite will just do the normal ALPB including non electrostatics. For the pure electrostatics, we have this new `--gbe` flag, which we also have in tblite.

src/tblite/calculator.F90
thfroitzheim: Yes, I agree but it turns our that the value is slightly different from the value one gets with the codata values for the fundamental constants (about 3e-12). That wouldn't be so bad, but I would not like to change this just here (it is also in tblite).

src/prog/main.F90
thfroitzheim: I think the opposite way is what we should aim for. Rewrite a clean propery module for tblite and then retire the main_property routine.

src/tblite/calculator.F90
thfroitzheim: It is actually a bit unclear what it is. The COSMO from xtb and the CPCM from tblite are the same code. From the epsilon scaling function it is a mixture of COSMO and CPCM. I would keep COSMO in xtb as a consistent naming and with the update to ddX in tblite, the problem should disappear.

src/tblite/calculator.F90
thfroitzheim: You are right that it doesn't make much sense to add a full function for this. I do the reading now manually. The reason is that we need to call the tblite internal solvent library, which I don't want to do in main. Once tblite is mandatory, we can change this.

PR #1401, opened 2026-04-27
Fix GFNFF SHAKE handling

Ehsanam1773:


thfroitzheim:
Please do not change the layout of the code, as this will otherwise be very hard to compare in the Github diff.

foxtran:
@thfroitzheim, 

there is an option to disable whitespace diff:

image

@Ehsanam1773, consider to split format changes and source code changes is different PRs.

PR #1413, opened 2026-05-31
Improve CMake reuse of vendored subprojects

Growl1234:
- Fix CMake target reuse so that subprojects already introduced by nested FetchContent/subproject builds are not added again, avoiding duplicate targets such as `test-drive`, `test-drive-lib`, and related test executables.
- Add Meson dist symlinks, so the CMake build can reuse the dependency sources already included in the dist tarball instead of downloading another copy online.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,14 +30,13 @@ project(
   LANGUAGES "C" "Fortran"
 )
 
-enable_testing()
-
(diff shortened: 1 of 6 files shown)


Growl1234:
👀 Gentle ping :-)

thfroitzheim (approved):

PR #1416, opened 2026-06-23
Fix and modernize the dormant GPU/cuSolver build

Prasanna163:
## What

The optional GPU code path (`WITH_GPU` / cuSolver, originally added in 2020) had
bit-rotted and no longer configures or compiles. This PR restores the build
plumbing **without touching the default CPU build** — all GPU bits stay guarded.

## Changes

- **`src/mctc/lapack/eigensolve.F90`** — fix an invalid local declaration
  `real(dp) :: dummy(:)` used for the cuSolver `bufferSize` query (a local
  assumed-shape array without `allocatable`/dummy status is not valid Fortran and
  broke the `USE_CUSOLVER` build). Replaced with `dummy(1)`, a valid placeholder
  for a size-only query.
- **`meson/meson.build`** — modernize deprecated NVIDIA HPC SDK flags for current
  `nvfortran`: `-ta=tesla:ccNN` → `-gpu=ccNN -acc`, `-Mcudalib` → `-cudalib`.
- **`CMakeLists.txt`** — add `WITH_GPU` / `GPU_ARCH` / `WITH_CUSOLVER` options
  (CMake had no GPU support at all), mirroring the meson logic; Fortran-scoped
  flags; defines `XTB_GPU` / `USE_CUSOLVER` / `USE_CUBLAS`.

## Testing

- The default CPU build (meson + gfortran) and the unit suite are unaffected —
  every change is behind the (off-by-default) GPU options.
- The cuSolver eigensolver path these flags enable was validated standalone on an
  RTX 3050: `cusolverDnDsygvd` vs LAPACK `dsygvd` agree to ~1e-14 eV on real
  GFN0 H/S matrices.

> Heads-up for context: a *full* GPU build of xtb is currently blocked further
> upstream by `nvfortran` internal compiler errors on the modern-Fortran
> dependency tree (deferred-length character handling, first seen in `toml-f`).
> This PR is just the build-system groundwork and is independent of that issue —
> the default build is unchanged either way. Happy to adjust anything.

### DIFF
--- a/src/mctc/lapack/eigensolve.F90
+++ b/src/mctc/lapack/eigensolve.F90
@@ -92,7 +92,7 @@ subroutine initDEigenSolver(self, env, bmat)
    integer :: istat, lwork
    ! dummy is only a dummy argument used to query the workspace size needed
    ! for cuSolverDnDsygvd -- it is okay to pass an empty array to cuSolverDnDsygvd_bufferSize
-   real(dp) :: dummy(:) 
+   real(dp) :: dummy(1)
 #endif
 
    self%n = size(bmat, 1)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,45 @@ enable_testing()
 # Follow GNU conventions for installing directories
 include(GNUInstallDirs)
 
+# ---------------------------------------------------------------------------
+# GPU acceleration (NVIDIA HPC SDK / nvfortran only).
+# Mirrors the meson options `gpu`, `gpu_arch`, `cusolver`. Disabled by default;
+# the CPU build is unaffected. Requires the compiler to be NVHPC (nvfortran).
+# Enable with: cmake -DWITH_GPU=ON -DGPU_ARCH=80 -DWITH_CUSOLVER=ON ...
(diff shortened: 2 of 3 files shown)


src/mctc/lapack/eigensolve.F90
```diff
@@ -90,9 +90,11 @@ subroutine initDEigenSolver(self, env, bmat)
...
 #ifdef USE_CUSOLVER
    integer :: istat, lwork
-   ! dummy is only a dummy argument used to query the workspace size needed
-   ! for cuSolverDnDsygvd -- it is okay to pass an empty array to cuSolverDnDsygvd_bufferSize
-   real(dp) :: dummy(:) 
+   ! dummy is only used to query the workspace size needed for cuSolverDnDsygvd;
+   ! the array contents are not dereferenced by the bufferSize query, a size-1
+   ! placeholder is sufficient. (Was previously declared `real(dp) :: dummy(:)`,
+   ! which is not valid Fortran for a local variable and broke the GPU build.)
+   real(dp) :: dummy(1)
```
foxtran: You can leave previous comment

PR #1418, opened 2026-07-04
GPU acceleration for GFN0/1/2 SCF, gradient and AES via a CUDA-C cuSolver/cuBLAS shim

Prasanna163:
## Summary

Adds optional GPU acceleration for GFN0/1/2 single points, gradients and
`--opt`, plus a native Windows build path. Real-world speedups measured on
an RTX 3050 (6GB): ~2x on a real protein (crambin, 642 atoms, SCF-convergence-
bound) up to ~5.9x-32x on larger/synthetic systems with more favorable
HOMO-LUMO gaps. Bit-exact vs CPU throughout (energies match to ~1e-15 Eh,
gradients to ~1e-16).

**Why this approach:** the natural route (building all of xtb with
`nvfortran` for OpenACC/CUDA Fortran) is blocked — nvfortran 26.3 ICEs on
sizing deferred-length `character(:)` in the dependency tree (toml-f etc.),
so the full xtb dependency stack can't currently compile with nvfortran.
Instead, `src/gpu/gpu_eig.cu` is a self-contained CUDA-C translation unit
(cuSolver + cuBLAS, compiled with `nvcc`) linked into the normal
gfortran-built `xtb` via `iso_c_binding`, gated behind `-Dgpu_shim=true`
and the `--gpu` CLI flag. Everything is opt-in and off by default; the
existing CPU path is untouched.

**What's included** (see individual commits for detail):
- CUDA-C cuSolver eigensolver shim + `--gpu-batch` cross-molecule GFN0
  batch driver (bucketed by AO size, one persistent CUDA context)
- Phase 2: GPU-resident GFN1/GFN2 SCF (S/H0/density kept on-device across
  the whole SCF, not re-uploaded per iteration)
- GPU analytical gradient kernel (`build_dSDQH0`: overlap/dipole/quadrupole
  integral-derivative contraction) for `--grad`/`--opt`
- GPU GFN2 anisotropic electrostatics (AES) kernels — opt-in via
  `XTB_GPU_AES=1` (default off: at pocket/protein scale the 8-thread CPU
  path is actually 3-10x faster than these particular kernels, which are
  memory/atomic-bound rather than compute-bound; left available for
  larger systems / future tuning)
- **Native Windows build** (this session's addition): since `nvcc` on
  Windows only accepts MSVC as its host compiler, its object files can't
  be linked directly into a MinGW/gfortran build. The shim is instead
  prebuilt out-of-band as a DLL (`nvcc`+MSVC), and a MinGW-compatible
  import library is generated from it via `dlltool`, which `xtb_exe`
  links against when `host_machine.system() == 'windows'`. The existing
  Linux `nvcc` custom_target path is untouched.
- Bug fix (`101b3a0`): `xtbopt.xyz` element-symbol corruption for atoms
  2..N was build-wide (all GFN methods), not GFN-FF-specific as originally
  suspected — root-caused to `molecule_to_structure` passing an
  oversized `character(len=80)` symbol array into mctc-lib.
- Dev tooling: benchmark scripts (`tools/bench/`), a GPU validation gate
  (`test/gpu/gfn12_gpu_gate.py`, wired into `meson test`), and Windows
  wrapper scripts (`tools/win/`) that auto-route small/large molecules to
  CPU/GPU.

**Known limitations / open questions for reviewers:**
- New Windows-only options (`gpu_shim_win_dir`, `gpu_cuda_bin_dir`) and
  the existing `nvhpc_root` option default to this developer's local
  paths, matching existing precedent for `nvhpc_root` in this file —
  happy to generalize/document these further if that's a blocker.
- GPU-vs-GPU thread-safety across molecules has only been validated for
  the H/C/N/O element set in the `--gpu-batch` path.
- This is a large, self-contained feature (adds a CUDA/cuSolver/cuBLAS
  dependency, opt-in via `-Dgpu_shim=true`). Happy to split into smaller
  reviewable pieces, or discuss design before merging, whatever's more
  useful.

## Test plan
- [x] `meson test` (existing xtb unit suite) passes unmodified on both
      the CPU-only and `-Dgpu_shim=true` builds
- [x] `test/gpu/gfn12_gpu_gate.py`: GFN1/GFN2 single-point, gradient and
      `--opt` bit-exact vs CPU (ΔE ~1e-15, |Δgrad| ~1e-16)
- [x] Real-protein validation (crambin, PDB 1CRN, 642 atoms): GPU vs CPU
      energies bit-identical, ~2x speedup (SCF-convergence-bound at this
      gap size)
- [x] Native Windows build (MSYS2/MinGW + MSVC/nvcc): CPU and GPU builds
      both validated bit-exact vs the Linux build on the same fixtures


foxtran:
That's nice to use LLMs, but, please, clean the stuff, that LLM generated.

Also, it would be better to split the whole work on small parts: no one wants to check 6k LoC by their eyes.

PR #1423, opened 2026-08-06
Fix rotational mode detection for diatomics

### LINKED ISSUE #1203: Segmentation fault for scan / constrained optimization for diatomic molecules

TyBalduf:
Fixes #1203

The `detrotra4/8` routines would score any mode with an eigenvalue below 0.05 to determine how well they preserved atomic distances, in order to classify low modes as translations/rotations vs vibrations. It would then sort these modes by score using the `ind` array and zero out the Hessian eigenvalues for the 6 (5 if linear) lowest scoring modes. 

This generally works fine, but if there aren't enough modes with an eigenvalue below 0.05, it would still try to zero out 5/6 modes even though some of the entries in `ind` would be undefined, leading to potential out of bounds access.

To avoid this, we check the number of low eigenvalue modes up front. If it isn't at least 5/6, we then check all the modes which should ensure we find 5/6 to remove. Checking all the modes shouldn't hurt performance, as we should only have this few small eigenvalues for small molecules, namely the diatomic case where a single value over the arbitrary 0.05 threshold was enough to cause a segfault.

-----

Beyond the bug fix, I also consolidated the interfaces for `detrota4/8` and have them use a common worker routine, since they perform the same algorithm just with single/double precision arrays.

### DIFF
--- a/src/detrotra.f90
+++ b/src/detrotra.f90
@@ -71,7 +71,7 @@ subroutine detrotra_worker(linear, n, xyz, h, eig, rigid)
    integer, allocatable :: ind(:)
    real(wp), allocatable :: mode(:), tmpxyz(:, :), score(:)
    real(wp) :: a0, b0, c0
-   logical :: check_all
+   logical, allocatable :: candidate(:)
 
    n3 = 3*n
    nend = min(merge(5, 6, linear), n3)
@@ -79,14 +79,18 @@ subroutine detrotra_worker(linear, n, xyz, h, eig, rigid)
    if (nend == 0) return
 
    ! Preserve the historic low-mode candidate set when it is large enough.
-   ! If it is not, inspect all modes instead of indexing uninitialized entries.
-   check_all = count(eig <= low_mode_threshold) < nend
+   ! Otherwise, add the remaining modes in ascending eigenvalue order until
+   ! there are enough candidates to identify all rigid-body modes.
+   allocate(candidate(n3), source=eig <= low_mode_threshold)
+   do while (count(candidate) < nend)
+      ii = minloc(eig, dim=1, mask=.not.candidate)
+      candidate(ii) = .true.
+   end do
    allocate(mode(n3), tmpxyz(3, n), score(n3), ind(n3))
 
    nn = 0
    do ii = 1, n3
-      ! Check only low-lying modes unless the full fallback is required.
-      if (.not.check_all .and. eig(ii) > low_mode_threshold) cycle
+      if (.not.candidate(ii)) cycle
 
       select type(h)
       type is(real(sp))
--- a/src/detrotra.f90
+++ b/src/detrotra.f90
@@ -20,12 +20,17 @@ module xtb_detrotra
    implicit none
    private
 
-   public :: detrotra4, detrotra8
+   public :: detrotra
+
+   interface detrotra
+      module procedure detrotra_sp
+      module procedure detrotra_wp
+   end interface detrotra
 
 contains
 
 !> Determine rotational and translational modes using single precision modes
-subroutine detrotra4(linear, n, xyz, h, eig)
+subroutine detrotra_sp(linear, n, xyz, h, eig)
    !> Whether the molecular structure is linear
    logical, intent(in) :: linear
    !> Number of atoms
@@ -41,10 +46,10 @@ subroutine detrotra4(linear, n, xyz, h, eig)
    call detrotra_worker(linear, n, xyz, h, real(eig, wp), rigid)
    ! Identifier for rotational and translational modes
    eig(rigid) = 0.0_sp
-end subroutine detrotra4
+end subroutine detrotra_sp
 
 !> Determine rotational and translational modes using double precision modes
-subroutine detrotra8(linear, n, xyz, h, eig)
+subroutine detrotra_wp(linear, n, xyz, h, eig)
    !> Whether the molecular structure is linear
    logical, intent(in) :: linear
    !> Number of atoms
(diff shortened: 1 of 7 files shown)


test/unit/test_detrotra.f90
```diff
@@ -0,0 +1,130 @@
+! This file is part of xtb.
+! SPDX-Identifier: LGPL-3.0-or-later
+!
```
lmseidler, line 3: probably should have a copyright header

test/unit/test_detrotra.f90
```diff
@@ -0,0 +1,130 @@
...
+   end interface build_diatomic_modes
+
+contains
+
+!> Collect all exported unit tests
+subroutine collect_detrotra(testsuite)
+   !> Collection of tests
+   type(unittest_type), allocatable, intent(out) :: testsuite(:)
+
+   testsuite = [ &
```
lmseidler, line 35: unit test naming should follow the repo convention (i.e. no whitespaces etc)

test/unit/test_detrotra.f90
```diff
@@ -0,0 +1,130 @@
...
+      new_unittest("single precision linear diatomic", test_detrotra4_diatomic), &
+      new_unittest("double precision linear diatomic", test_detrotra8_diatomic), &
+      new_unittest("preserve sufficient low-mode candidates", test_low_mode_candidates) &
+      ]
+
+end subroutine collect_detrotra
+
```
lmseidler, line 43: error thr in terms of epsilon(0.0_sp) and epsilon(0.0_wp) respectively. also, 1e-5 seems a bit large?

test/unit/test_detrotra.f90
```diff
@@ -0,0 +1,130 @@
...
+   eig = [0.01_wp, 0.01_wp, 0.01_wp, 0.01_wp, 0.01_wp, 0.01_wp, &
+      3.0_wp, 4.0_wp, 5.0_wp]
+
+   call detrotra8(.false., nat, xyz, hess, eig)
+
+   call check(error, count(eig == 0.0_wp), 6)
+   call check(error, eig(7), 3.0_wp, thr=1.0e-12_wp)
+end subroutine test_low_mode_candidates
+
+subroutine build_diatomic_modes_sp(hess)
```
lmseidler: these helpers can be removed and the hessians hardcoded as fixtures in the tests directly.

src/detrotra.f90
```diff
@@ -16,132 +16,111 @@
...
-   real(wp)              :: a0,b0,c0
+   private
+
+   public :: detrotra4, detrotra8
+
+contains
+
+!> Determine rotational and translational modes using single precision modes
+subroutine detrotra4(linear, n, xyz, h, eig)
+   logical, intent(in) :: linear
```
lmseidler, line 35: linear and n should also have docstrings if the other variables have them too.

src/detrotra.f90
```diff
@@ -16,132 +16,111 @@
...
+
+      ! Distort the initial geometry along the ii-th mode.
+      tmpxyz = xyz + reshape(mode, shape(tmpxyz))
+
+      ! Compare all interatomic distances of the original and distorted
+      ! geometries.
+      c0 = 0.0_wp
+      do i = 2, n
+         do j = 1, i - 1
+            a0 = sqrt(sum((xyz(:, i) - xyz(:, j))**2))
```
lmseidler: this should be just norm2 instead of the comprehension, same below

lmseidler (changes requested):

lmseidler (approved):

thfroitzheim (approved):
Thank you, this is a clear improvement over always checking only 5/6 eigenvalues. Anyway, we should think long-term about switching to projecting out rotations and translations instead of trying to assign them to the full Hessian (we already have code for the projection [here](https://github.com/grimme-lab/xtb/blob/main/src/freq/project.f90)).

PR #1424, opened 2026-08-06
Add `$seed <int>` xcontrol keyword

### LINKED ISSUE #730: Provide a seed parameter for the random number generator
Sometimes it is desired to generate MD trajectories in a reproducible manner, but there currently seems no way to provide a custom seed for the random number generator.

Provide a command line option like ```--seed``` or ```xcontrol``` parameter ```$seed```

**Describe alternatives you've considered**
A current workaround is to generate an initial trajectory and then reuse the corresponding ```mdrestart``` file.

**If possible state how you can assist in providing data or code to to implement the feature**
---

**Additional context**
---

### LINKED ISSUE #1247: External seed for metadynamics
For reproducibility, it would be nice to have a way to set external seed. At the current moment, only quite strange procedure `initrand` exists:
https://github.com/grimme-lab/xtb/blob/04ad2f85c501df22c555749200791e13d5f8883b/src/setparam.f90#L535-L540
which allows to use the same seed between all runs without an ability to change it without recompilation.

TyBalduf:
Allow a user to set a specific random seed, rather than the specific value of 41 supplied by `samerand`. We now print the random seed, so this enables a user to exactly reproduce a specific metadynamics run (given the same compiler, threading, etc).

Different Fortran compilers require a different number of seed values, M. We now pass the auto generated or user selected seed through a simple random number generator. We generate M values and use these as the seeds for the production random number generation.

Fixes #1247, Fixes #730


src/set_module.f90
```diff
@@ -1125,11 +1131,36 @@ subroutine set_enso_mode
...
 end subroutine set_enso_mode
 
+!> Request the legacy repeatable sequence, now represented by scalar seed 41.
 subroutine set_samerand
    implicit none
    set%samerand = .true.
 end subroutine set_samerand
 
+!> Read the scalar RNG seed from a top-level `$seed <int>` instruction.
+!>
```
lmseidler: one-line comment header is enough

src/setparam.f90
```diff
@@ -538,24 +541,79 @@ module xtb_setparam
 contains
 
 
+!> Initialize the Fortran pseudorandom number generator from one scalar seed.
+!>
```
lmseidler: again one line. in general i would prefer comments to be short and relevant instead of explaining the function in the same number of lines as the actual function, this is more inline with the remaining source and more readable.

src/setparam.f90
```diff
@@ -538,24 +541,79 @@ module xtb_setparam
...
+         set%randseed = int(modulo(int(seed_vector(1), int64), seed_modulus - 1_int64))
+      endif
+      set%randseed_set = .true.
    endif
+
+   call expand_random_seed(set%randseed, seed_vector)
+   call random_seed(put=seed_vector)
 end subroutine initrand
 
+!> Expand one user-facing seed into the processor-dependent seed vector.
```
lmseidler: same as above. also: this is another case where you can just merge this into the caller since it is not used anywhere else and only called once. the @... in the docstring is not used most of the time in our code afaik. instead `!>` docstrings above the variable declarations are common.

test/unit/test_random.f90
```diff
@@ -0,0 +1,167 @@
+! This file is part of xtb.
+! SPDX-Identifier: LGPL-3.0-or-later
+!
```
lmseidler, line 6: this should probably have a copyright header

test/unit/test_random.f90
```diff
@@ -0,0 +1,167 @@
...
+   set%randseed_set = .true.
+   set%samerand = .false.
+   call initrand
+   call random_number(seed_values)
+   call check(error, all(samerand_values == seed_values))
+
+   call restore_random_state(previous_state)
+end subroutine test_samerand
+
+!> Capture all global state changed by the random-seed unit tests.
```
lmseidler: the comments from above similarly apply here

lmseidler (changes requested):
Thanks again for your contributions. Only some minor changes, mostly docstring stuff.

lmseidler (approved):

PR #929, opened 2023-12-12
Periodic boundary conditions in GFN-FF

Thomas3R:
Periodic boundary conditions are now available in xtb when using either the `--gfnff` or `--mcgfnff` command line argument. Here, GFN-FF refers to the original parameterization, and mcGFN-FF to a version with scaled non-covalent interactions for molecular crystals. As with other methods, the lattice parameters have to be present in the input file in order for xtb to recognize a periodic system.

PR #1425, opened 2026-08-06
Fix gfnff topology restoration

### LINKED ISSUE #987: Error while trying to convert 2d structure to 3d using gfn-ff
**Describe the bug**
Hi, I encountered a problem when trying to use gfn-ff to convert a two-dimensional molecular structure into 3D. For some reason, xtb is unable to read the connectivity information of atoms from the file. The molecule file is valid and opens with other programs (ketcher, marvin sketch, ...). I am attaching the molecule file and xtb output to the email.Different file formats cause xtb to behave differently. When using the .mol format, the program crashes with an error, but if the same structure is converted to .sdf, the program runs and optimization results in all atoms flying apart at huge distances from each other.
Could this be a bug, or am I preparing the data incorrectly for processing?

**To Reproduce**
xtb C\(C\)\(P\(S\)\(\=S\)C\(C\)CCCCCC\)CCCCCC_ketcher.sdf --gfn 2 --sp
[mol_file.txt](https://github.com/grimme-lab/xtb/files/14543028/mol_file.txt)

Output files
[xtb_output_mol.txt](https://github.com/grimme-lab/xtb/files/14543055/xtb_output_mol.txt)
[xtb_output_sdf.txt](https://github.com/grimme-lab/xtb/files/14543056/xtb_output_sdf.txt)

All calculations were carried out on laptop with Arch Linux using xtb 6.7.0

**Expected behaviour** 
Valid transformation of 2D coords to 3D

### LINKED ISSUE #1132: xtb dock geometry "exploding" during optimization using GFN-FF
**Describe the bug**
During the optimization step, the geometry "explodes" when using GFN-FF, i believe because there is no topology generated - or not as far as i can see.

**To Reproduce**

1. structure 1 = receptor_xyz.txt [ligand_xyz.txt](https://github.com/user-attachments/files/17879188/ligand_xyz.txt) 
structure 2 = ligand_xyz.txt [receptor_xyz.txt](https://github.com/user-attachments/files/17879189/receptor_xyz.txt)
part of optimization of first pose using the pocket search = pocket_opt_example_xyz.txt [pocket_opt_example_xyz.txt](https://github.com/user-attachments/files/17879202/pocket_opt_example_xyz.txt)

2. xtb command:
`xtb dock receptor.xyz ligand.xyz --optlvl gfnff --opt sloppy --alpb water --pocket --atm &> xtb_docking.out`

Please provide all input and output file such that we confirm your report.

**Expected behaviour**
The topology of structure 1 and structure 2 should be preserved. 

**Additional context**
This does not happen when selecting, e.g., `--optlvl gfn1` so is specific to applying the gfn-ff method. This, however, runs a lot slower.

TyBalduf:
Fixes #987, Fixes #1132 

#987 was introduced by https://github.com/grimme-lab/xtb/pull/929, where (among other changes) `TGFFTopology` `blist` and `nbond` attributes were no longer set, with most usages replaced by the corresponding attributes of `TNeigh`. The 2D->3D conversion code missed this remapping, so the harmonic restraints were using faulty `topo%nbond` and `topo%blist` rather than the up to date `neigh%nbond` and `neigh%blist` as intended. Updating those resolves the issue.

#1132 stems from artificial bonds between the ligand and receptor being formed during close contacts. Some of these "bonds" would be nonsensical like an H-H bond. The proposed fix is to restore the separated bonding for ligand and receptor, just allowing them to interact non-covalently.

### DIFF
--- a/src/docking/search_nci.f90
+++ b/src/docking/search_nci.f90
@@ -59,7 +59,7 @@ module xtb_docking_search_nci
    implicit none
 
    private
-   public :: docking_search
+   public :: docking_search, restore_gff
 
 contains
 
@@ -544,14 +544,7 @@ contains
             !> Select which kind of optimization is done
             select type (calc)
             type is (TGFFCalculator)
-               call restart_gff(env, comb, calc)
-               !Keeping Fragments and charges
-               calc%neigh%nbond = neigh_backup%nbond
-               calc%neigh%nb = neigh_backup%nb
-               calc%topo%qfrag = topo_backup%qfrag
-               calc%topo%qa = topo_backup%qa
-               calc%topo%fraglist = topo_backup%fraglist
-               calc%topo%nfrag = topo_backup%nfrag
+               call restore_gff(env, comb, calc, topo_backup, neigh_backup)
             type is (TxTBCalculator)
                call restart_xTB(env, comb, chk, calc)
             end select
@@ -916,13 +909,7 @@ contains
 
          select type (calc)
          type is (TGFFCalculator)
-            call restart_gff(env, comb, calc)
-            calc%neigh%nbond = neigh_backup%nbond
-            calc%neigh%nb = neigh_backup%nb
-            calc%topo%qfrag = topo_backup%qfrag
-            calc%topo%qa = topo_backup%qa
-            calc%topo%fraglist = topo_backup%fraglist
-            calc%topo%nfrag = topo_backup%nfrag
+            call restore_gff(env, comb, calc, topo_backup, neigh_backup)
          type is (TxTBCalculator)
             call restart_xTB(env, comb, chk, calc)
          end select
@@ -1073,12 +1060,7 @@ contains
 
       integer :: itopo = 32
 
-      !> Read the constrain again with new xyz only if necessary
-      if (constraint_xyz) then
-         nconstr=0 !Reset number of constraints for distance, angle, and dihedral
-         call read_userdata(xcontrol, env, mol)
-         call constrain_xTB_gff(env, mol)
-      end if
+      call update_gff_constraints(env, mol)
 
       !if(auto_walls) call number_walls=0; read_userdata(xcontrol,env,mol) !everytime new wall pot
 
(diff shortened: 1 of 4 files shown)


src/docking/search_nci.f90
```diff
@@ -1109,6 +1091,64 @@ subroutine restart_gff(env, mol, calc)
...
+   end subroutine restore_gff
+
+   !> @brief Rebuild coordinate-dependent GFN-FF constraints for a new geometry.
+   !>
+   !> User constraints whose values are taken from the input coordinates must be
+   !> reread whenever docking changes the molecule or pose.  Reset the global
+   !> distance, angle, and dihedral constraint count before parsing to avoid
+   !> accumulating entries from the previous geometry.  If no coordinate-based
+   !> constraints were requested, this routine is a no-op.
+   subroutine update_gff_constraints(env, mol)
```
lmseidler: i think this can be merged into the restore_gff subroutine, not used anywhere else and is already pretty short. i would also say the comment headers could be a bit less verbose.

src/docking/search_nci.f90
```diff
@@ -59,7 +59,7 @@ module xtb_docking_search_nci
    implicit none
 
    private
-   public :: docking_search
+   public :: docking_search, restore_gff
```
lmseidler: this shouldn't be public. it is not called anywhere else except the restore topology test. i would also say the test is pretty trivial so could probably be left out.

test/unit/test_docking.f90
```diff
@@ -641,4 +642,88 @@ subroutine test_dock_wat_wat_gfnff(error)
 
 end subroutine test_dock_wat_wat_gfnff
 
+
+subroutine test_restore_gfnff_topology(error)
```
lmseidler: if this test stays it should match the patterns for using molecule fixtures from other tests (use molstock or mstore, i think in xtb it should be molstock).

test/unit/test_gfnff.f90
```diff
@@ -131,6 +132,53 @@ subroutine test_gfnff_sp(error)
...
+   call check_(error, .not.allocated(calc%topo%blist))
+
+   call calc%singlepoint(env, mol, chk, 0, .false., energy, gradient, sigma, &
+      & hl_gap, res)
+
+   call env%check(exitRun)
+   call check_(error, .not.exitRun)
+   call check_(error, ieee_is_finite(energy))
+   call check_(error, all(ieee_is_finite(gradient)))
+   call check_(error, energy, 0.00476278587765942_wp, thr=1.0e-12_wp)
```
lmseidler: thr should be in terms of epsilon. also does this test against regression from before the change or specifically the failure case?

lmseidler (changes requested):

test/unit/test_gfnff.f90
TyBalduf: Yes, this checks not only that we get the right value for the failed case, but also that the correct `blist` array is allocated/used.

lmseidler: i mean, can the energy value be reproduced with the old code?

TyBalduf: Oh sorry, yes the energy/gradient values agree to machine precision with the values from pre #929. Prior to the current fix, the energy value was off by~ 0.003 hartree and the gradient norm was off by ~0.04 hartree/bohr.

lmseidler (approved):

PR #1426, opened 2026-08-07
Remove molecular rotation before numerical Hessian

### LINKED ISSUE #1227: Slight rotation of molecule before computing Hessian
This might not be a bug, but I couldn't find an explanation for why this was done. Before computing the numerical Hessian, several small rotations are applied to the molecule: https://github.com/grimme-lab/xtb/blob/41059b44d84a07dfd99a051e4409ef654faca939/src/hessian.F90#L143.

This leads to (small) deviations in comparison to manual calculation of the numeric Hessian or computing it via the C-API. Is there a reason this is done?

TyBalduf:
* While tiny, this causes direct comparison with an externally computed numerical Hessian (or the Hessian reported by the C-API) not to match.

Fixes #1227

### DIFF
--- a/src/hessian.F90
+++ b/src/hessian.F90
@@ -107,7 +107,6 @@ subroutine numhess( &
    real(wp),allocatable :: aux (:)
    real(wp),allocatable :: isqm(:)
    real(wp),allocatable :: gl  (:,:)
-   real(wp),allocatable :: xyzsave(:,:)
    real(wp),allocatable :: pold(:)
    real(wp),allocatable :: dipd(:,:), dalphadr(:,:), dalphadq(:,:)
    real(wp),allocatable :: amass_au(:), amass_amu(:)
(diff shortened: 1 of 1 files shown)


lmseidler:
Thanks for your contributions. Looks good. Are you aware of any possible reason why this was done? I was trying to figure that out myself for some time now...

TyBalduf:
@lmseidler I couldn't figure out why this was the present. It doesn't have a clear physical reason and I know from comparison that other QM programs like Jaguar and Psi4 don't do this. 

My best guess is that in earlier xTB versions (farther back then we can see in the git history), they were working around some issue with computing the Hessian for axis aligned structures. And because it was never documented why this was present or what it was trying to accomplish, it was just carried forward with any updates to this file.

This doesn't need to get merged anytime soon if there are some concerns that the rotation has a valid use case. I just figured I'd put up the PR to highlight it a bit more than the old issue.

lmseidler (approved):

Issue #1333, opened 2025-08-06
xtb thermo --turbomole


debruinb:
I think I found a bug in "xtb thermo --turbomole" in analysing externally generated tubomole hessian files. 
It seems to generate a wrong (way too large) ZPE correction.  

**To Reproduce**
Using: xtb 6.7.1

[Turbomole.zip](https://github.com/user-attachments/files/21616843/Turbomole.zip)

**xtb thermo --turbomole coord hessian > thermo.out** 
(generates way too large zpe using the files attached: 9.575952807020 Eh) 

**xtb thermo coord hessian >thermo.out**
(generates the correct zpe (0.123974373523 Eh) corresponding roughly to the output of aoforce (0.1239725  Hartree) ) 

**Expected behaviour**
correct zpe correction term, corresponding to aoforce output 

**Additional context**
Handy tool to analyze thermodynamic data of turbomole, but it seems the --turbomole flag reveals a bug

thfroitzheim:
As described in #1427, this happens because the Hessian file is assumed to already be mass-weighted (contrary to the previous help message). Now, mass-weighting is not assumed for a Turbomole input but separately activated.

PR #1427, opened 2026-08-07
Thermo submodule: make mass weighting a seperate keyword

chselz:
This change separates now the input hessian format from the mass-weighting boolean in the thermo submodule. This lead in the past to confusion with issue #1333 which should be closed by this PR

### DIFF
--- a/src/prog/thermo.f90
+++ b/src/prog/thermo.f90
@@ -270,6 +270,8 @@ contains
 
          case ('--turbomole')
             htype = hessType%tmol
+
+         case ('--mass-weighted', '--massweighted')
             massWeighted = .true.
 
          case ('--orca')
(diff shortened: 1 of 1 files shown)


src/prog/thermo.f90
```diff
@@ -373,7 +375,8 @@ subroutine thermoHelp(unit)
          "   --dftbplus             Read a DFTB+ hessian.out file, implies projection", &
          "", &
          "   --turbomole         Read a Turbomole Hessian file", &
-         "                       use this only when $nomw is not present in control", &
```
lmseidler, line 376: i'm not sure if it is a good idea to delete this, it should probably say "Read a non-mass-weighted Turbomole Hessian file" (repeats stuff a bit but better safe than sorry). also for `--mass-weighted` i would write instead: "Assume mass-weighted input Hessian".

lmseidler (changes requested):

lmseidler (approved):

PR #1433, opened 2026-08-09
Update tblite dependency to v0.7.0

thfroitzheim:
This PR updates the tblite dependency to v0.7.0. This adds support for: 
- ddX solvation models (COSMO, CPCM, and PCM)
- Molden file output with tblite
- a more complete output, e.g., for multipole moments with tblite

I also tightened the SCF convergence accuracy consistently with the optimization level and for numerical Hessians. In particular, the tighter settings for the Hessian should remove some of the asymmetry often observed with the numerical Hessian. 

Lastly, I added the results object to be updated during the final singlepoint in the geometry optimization. Otherwise, the MO coefficients in the Molden file don't belong to the final geometry (Issue reported at [g-xtb#34](https://github.com/grimme-lab/g-xtb/issues/34)). 

I noticed a problem with the mctc-lib version 0.5.2 and the CMake build tree export (currently deactivated) and the Windows build has some problems with ddX. 

@robin-dahl


Growl1234:
Regarding DDX failure, I am having a PR to allow DDX to be built as static library (https://github.com/ddsolvation/ddX/pull/182).

Growl1234 (commented):
It's suggested to add
```bash
link_subproject tblite ddx
```
to [xtb/meson/dist_symlinks.sh](https://github.com/grimme-lab/xtb/blob/a45f05d397268fd683b5f9f824ed59912c5490b5/meson/dist_symlinks.sh#L51-L57).

Growl1234:
Sorry, but I cannot understand what b1b1c38 does.

awvwgk:
@robin-dahl Looks like you committed your local CMake build cache files to the repo.

robin-dahl:
@awvwgk Yes, that was not intentional, sorry about that! 

The ddX Windows build, however, affects both the regular and static Meson builds. ddX unfortunately adds its internal test executables to meson's default build graph when it is included through tblite. One of these executables (`ddx_driver_testing`) uses a Fortran source file as a symlink. On the Windows runner, that symlink is treated as a regular file which GFortran fails to compile of course.

The simplest solution would be to disable ddX's internal test suite when ddX is consumed as a subproject. However, currently there is not meson option for doing so in the latest ddX release version. Thus, there is no absolute clean way to solve this issue in xTB without adding a small change to the ddX subproject itself.

awvwgk:
Handling of ddx via meson at the moment is quite suboptimal, I think we have to disable it by default in tblite if support is not improved upstream or an alternative way to discover it, e.g. via pkg-config, is made available. Some issues were already reported upstream: https://github.com/ddsolvation/ddX/issues/179 and https://github.com/ddsolvation/ddX/issues/180. I believe meson support in ddx is owned (or at least contributed) by the xtb team.

PR #1434, opened 2026-08-10
Fix file descriptor leak

### LINKED ISSUE #1388: File descriptor leak in C-API
**Describe the bug**

In src/api/environment.f90, the releaseOutput_api subroutine (bound as xtb_releaseOutput) has a logic error that prevents file units from being closed. This causes a file descriptor leak when the C-API is used repeatedly (e.g., in geometry optimisation or MD), eventually hitting the OS limit with "Too many open files".

On line 203 (commit fc512bb):
```
  if (.not.any(env%ptr%unit /= [-1, stdout, stderr])) then
     close(unit=env%ptr%unit, iostat=stat, iomsg=message)
```

the condition is always `.false.` by defining not any of `-1`, `stdout`, or `stderr`, therefore `close` is never triggered.

**To Reproduce**

Here is a minimal example:

```
#include "xtb.h"
#include <stdio.h>
void main() {
    for (int i = 0; i < 1200; i++) {
        xtb_TEnvironment env = xtb_newEnvironment();
        char filename[255];
        snprintf(filename, sizeof(filename), "xtb_step_%d.log", i);
        xtb_setOutput(env, filename);
        // ... run calculation ...
        xtb_delEnvironment(&env);  // within this the file unit isn't closed as expected
        printf("### Step %d completed\n", i);
    }
}
```

This will crash at around step ~1000 (depending on ulimit -n) with an error opening a new file.

**Expected behaviour**

`xtb_delEnvironment` which calls `releaseOutput_api` should close file units that are not `-1`, `stdout`, or `stderr`. It can be easily fixed by changing `/=` to `==`:

`if (.not.any(env%ptr%unit == [-1, stdout, stderr])) then`

**Additional context**

The bug was introduced in commit 37ae399a (C-API version 6.3, #222) in May 2020.

Thank you very much.

TyBalduf:
`releaseOutput_api` had a logic bug, where it would only attempt to free stdout, stderr, or an unset file descriptor, where the intended behavior was to free any other file descriptor.

Fixes #1388

### DIFF
--- a/src/api/environment.f90
+++ b/src/api/environment.f90
@@ -176,7 +176,7 @@ subroutine releaseOutput_api(venv) &
       call c_f_pointer(venv, env)
       call checkGlobalEnv
 
-      if (.not.any(env%ptr%unit /= [-1, stdout, stderr])) then
+      if (.not.any(env%ptr%unit == [-1, stdout, stderr])) then
          close(unit=env%ptr%unit, iostat=stat, iomsg=message)
          if (stat /= 0) then
             call env%ptr%error(trim(message), source)
--- a/test/api/c_api_example.c
+++ b/test/api/c_api_example.c
@@ -4,6 +4,11 @@
 #include <stdlib.h>
 #include <stdio.h>
 
+#if defined(__unix__) || defined(__APPLE__)
+#include <sys/resource.h>
+#include <unistd.h>
+#endif
+
 #include "xtb.h"
 
 #define check(x, ...)                                                          \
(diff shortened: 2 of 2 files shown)


TyBalduf:
@lmseidler @thfroitzheim sorry for the glut of PR reviews, but just happened to notice this issue, which I think is the source of crashes we have seen at Schrodinger on Windows when making heavy use of the C-API in tests.

I think with this change incorporated, Schrodinger would be able compile our next xTB version that would be usable across the platforms and compilers we support. Thank you! 

And thank you to @chemsh for reporting this and the suggested fix.

thfroitzheim (approved):
No worries, and sorry that the reviews are taking so long. Thank you for all the effort you put into fixing xtb!

PR #1435, opened 2026-08-11
Hessian Rot/Trans Projection after Mass-weighting

### LINKED ISSUE #693: linear water calculated to have no imaginary frequencies
**Describe the bug**
I constructed a linear water molecule as

```
3

H  0.0  0.0  1.0
O  0.0  0.0  0.0
H  0.0  0.0 -1.0    
```

and used xtb 6.5.1 to optimize it. I played around with some parameters as with or without gbsa, --acc and other parameters; I also looked if --ohess versus separate --opt and --hess calculations have an influence. They don't.

It results in a linear water molecule as expected but it calculates all positive frequencies, which surprised me.

**To Reproduce**

1. create water molecule input `h2o.xyz` as above
2. load xtb 6.5.1 module

```
module show xtb/6.5.1
-------------------------------------------------------------------
/data/software/modules/xtb/6.5.1:

setenv           XTBHOME /scratch/software/xtb-6.5.1 
setenv           XTBPATH /scratch/software/xtb-6.5.1/share/xtb 
prepend-path     MANPATH /scratch/software/xtb-6.5.1/share/man 
prepend-path     PATH /scratch/software/xtb-6.5.1/bin 
append-path      PKG_CONFIG_PATH /scratch/software/xtb-6.5.1/lib64/pkgconfig 
setenv           OMP_NUM_THREADS 9,1 
setenv           OMP_MAX_ACTIVE_LEVELS 1 
setenv           MKL_NUM_THREADS 9 
setenv           OMP_STACKSIZE 100G 
-------------------------------------------------------------------
```

4. run `xtb` with your options and the `--verbose` flag
`xtb h2o.xyz --acc 0.000001 --ohess extreme --alpb water --verbose &> xtb_opt.log`
6. output showing the error
[xtb_opt.log](https://github.com/grimme-lab/xtb/files/9498586/xtb_opt.log)

**Expected behaviour**
Linear water is supposed to have two imaginary/negative frequencies.

chselz:
Before the projection of the rot/trans modes were done before the mass weighting. In the dxtb project as well as the gaussian white paper reads as if the correct way to do this is, however, after the massweighting. this lead to problems as no projection was done for linear molecules resulting in wrong frequencies. This patch should elevate this problem, works also for fixed atoms. Testing on the in #693  mentioned linear water molecules yields now two -1802.38 cm^-1 modes. the ordering in the vibspectrum file also seems to be correct now.

This should therefore close #693 .


lmseidler:
could you explain a bit more: was the problem that no projection was done for linear molecules or that projection before mass weighting changed frequencies significantly in a non-linear case or both? and can you give an example to show both? it might make sense to pin relevant behavior in unit tests. also after merging #1446 this might need some slight updates (there is already a merge conflict)

PR #1436, opened 2026-08-12
Ensure a restart file is written after convergence failure

### LINKED ISSUE #1146: --restart not working and faulty convergence with wrong restart files
**Describe the bug**
The `--restart` flag does not create any restart files when a calculation is not converging.

**To Reproduce**

1. Happens with  [struc.zip](https://github.com/user-attachments/files/18096692/struc.zip) using `xtb --restart struc.xyz` using `xtb 6.7.1`. The failed convergence is an expected behavior.
2. Using `xtb 6.3.0` restart files can be created that can be read in and lead to convergence with faulty Wiberg bond orders and wrong topology.

TyBalduf:
Fixes #1146 

Just adds writing/dumping of restart files to before the calculation termination that would occur if the initial SCC failed to converged.

### DIFF
--- a/src/prog/main.F90
+++ b/src/prog/main.F90
@@ -1245,6 +1245,18 @@ contains
       end if
 
       ! ------------------------------------------------------------------------
+      !  to further speed up xtb calculations we dump our most important
+      !  quantities in a restart file, so we can save some precious seconds
+      select type (calc)
+      type is (TxTBCalculator)
+         if (restart) then
+            call writeRestart(env, chk%wfn, 'xtbrestart', set%gfn_method)
+         end if
+      type is (TTBLiteCalculator)
+         if (restart) call dumpRestart(env, chk, 'xtbrestart')
+      end select
+
+      ! ------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


src/prog/main.F90
```diff
@@ -1239,18 +1244,6 @@ subroutine xtbMain(env, argParser)
          call stop_timing(10)
       end if
 
-      ! ------------------------------------------------------------------------
-      !  to further speed up xtb calculations we dump our most important
-      !  quantities in a restart file, so we can save some precious seconds
```
thfroitzheim, line 1244: Is this still consistent, if one performed an optimization or MD in between? In this case, we want to write the restart file for the new structure, right? Maybe we have to leave this here

TyBalduf: Ah you are right, I was for some reason thinking it looped backed around to the single point, but this is just the initial single point done before any other calculation mode. I've restored this, so the only change is just an additional restart dump for the initial SCC incase it fails.

thfroitzheim (approved):

PR #1446, opened 2026-09-15
Implementation of compliance elements and reusable internal coordinate infrastructure

lmseidler:
The goal of this PR is to introduce the calculation of compliance elements from the Cartesian Hessian. In the process I'll also implement a robust internal coordinate implementation, which can later be used to replace redundant code in different modules (intmodes, geosum) in a separate PR. 

Edit: there is also potential to further reduce redundant code in model Hessian infrastructure but this needs some special internal coordinate generation rules and is beyond the scope of this PR. Added a TODO though.


src/internals/graph.f90
```diff
@@ -0,0 +1,143 @@
...
+!> spanning tree per connected component.
+module xtb_internals_graph
+   use xtb_type_neighbourlist, only : TNeighbourList
+   implicit none
+   private
+
+   public :: graph_type, init
+
+   !> Molecular graph of a system
+   type :: graph_type
```
thfroitzheim, line 33: This should likely follow the type naming convention in xtb so T...

src/internals/redundant.f90
```diff
@@ -0,0 +1,196 @@
...
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with xtb.  If not, see <https://www.gnu.org/licenses/>.
+
+!> Redundant internal coordinates of a system.
+!>
+!> The coordinate values are evaluated once, at the reference geometry handed
+!> to init, and stored in q. They are a snapshot and go stale when the atoms
+!> move; build the set again from the geometry of interest.
+module xtb_internals_redundant
```
thfroitzheim, line 23: Shouldn't redundant internal coordinates already be part of the geometry optimizer? Can this cleaner implementation be reused there?

src/internals/redundant.f90
```diff
@@ -0,0 +1,196 @@
...
+   !> Redundant internal coordinates of a system. The coordinate definitions
+   !> are the inherited ones, the counts and the reference values are the
+   !> redundant set's own state.
+   type, extends(internal_coords_set_type) :: redundant_type
+
+      !> Number of atoms in the system
+      integer :: n = 0
+
+      !> Number of bonds, angles and dihedrals, ncoords = nbond + nangle + ndihedral
+      integer :: nbond = 0, nangle = 0, ndihedral = 0
```
thfroitzheim, line 44: Separate rows

src/internals/redundant.f90
```diff
@@ -0,0 +1,196 @@
...
+!> Build the coordinate definitions from the molecular graph and evaluate them
+!> at the reference geometry.
+pure subroutine new_redundant(self, graph, xyz)
+   !> Coordinate set to fill, its allocatable components are allocated here
+   type(redundant_type), intent(out) :: self
+   !> Molecular graph of the system
+   type(graph_type), intent(in) :: graph
+   !> Cartesian reference coordinates, dimension (3, graph%n), in the length
+   !> unit the coordinate values are reported in
+   real(wp), intent(in) :: xyz(:, :)
```
thfroitzheim, line 68: Wouldn't it be better to take a structure object in here instead of moving around its components?

src/internals/redundant.f90
```diff
@@ -0,0 +1,196 @@
...
+   end do
+   do ia = self%nbond + self%nangle + 1, self%ncoords
+      self%q(ia) = dihedral_value(xyz(:, self%atoms(1, ia)), xyz(:, self%atoms(2, ia)), &
+         & xyz(:, self%atoms(3, ia)), xyz(:, self%atoms(4, ia)))
+   end do
+
+end subroutine new_redundant
+
+!> Distance between atoms i and j
+pure function dist(xyz, i, j) result(r)
```
thfroitzheim, line 172: Are separate routines really necessary if they are called only once?

src/internals/type.f90
```diff
@@ -0,0 +1,53 @@
...
+!> A coordinate set is a plain definition of which coordinates exist and
+!> which atoms they involve; the values of the coordinates are not part of
+!> it. Concrete sets extend this type and add their own state.
+module xtb_internals_type
+   implicit none
+   private
+
+   public :: internal_coords_set_type, coord_bond, coord_angle, coord_dihedral
+
+   !> Bond coordinate kind, one coordinate per edge
```
thfroitzheim, line 29: Should this be an enum?

src/model_hessian/gff.f90
```diff
@@ -284,6 +299,8 @@ end subroutine get_pair_hessian
 
 !> Evaluate a diagonal pair-Hessian element
 pure subroutine getqqxx(dx, qq, cdisp, r, r2, r3, damped_r, r0_squared, d2)
+   !> Cartesian separation component, Coulomb and dispersion pair factors,
```
thfroitzheim, line 302: Separate lines and commens

src/model_hessian/internal.f90
```diff
@@ -31,28 +31,38 @@ module xtb_modelhessian_internal
...
    use xtb_type_param, only : chrg_parameter
    use xtb_type_environment, only : TEnvironment
+   use xtb_type_setvar, only : modhess_setvar
    implicit none(type, external)
    private
 
    !> Base for model Hessians defined by redundant internal coordinates
    type, public, abstract, extends(TModelHessian) :: TInternalModelHessianBase
+      private
+      real(wp) :: kr, kf, kt, ko, kd, kq, rcut, s6
```
thfroitzheim, line 41: Comments what those are (also for the procedures)

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+!> angles, and dihedrals.
+!>
+!> Also calls write_compliance_dat to write the full matrix to a file.
+subroutine print_compl(unit, n, at, mass, internals, C)
+   !> Redundant coordinate set with the definitions and the reference values.
+   type(redundant_type), intent(in) :: internals
+   !> Formatted output unit.
+   integer, intent(in) :: unit
+   !> Number of atoms.
+   integer, intent(in) :: n
```
thfroitzheim, line 128: Can we make the naming consistent with the structure type (or again just put a structure in here)

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   !> Number of internal coordinates.
+   integer, intent(in) :: ncoord
+   !> Atomic numbers, dimension (n).
+   integer, intent(in) :: at(n)
+   !> Compliance matrix, dimension (ncoords, ncoords), in Bohr^2/Hartree.
+   real(wp), intent(in) :: C(internals%ncoords, internals%ncoords)
+   !> Coordinate order used for the printed table, dimension (ncoord).
+   integer, intent(in) :: idx_ord(ncoord)
+
+   integer, parameter :: NCOUP = 20
```
thfroitzheim, line 231: What is this?

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+
+!> Compliance constants of the reference geometry as the pseudoinverse of the
+!> Hessian projected out of translations and rotations.
+!>
+!> C = B H^+ B^T, evaluated as Z (Z D)^T with Z = B V and D = diag(1/w_i) from
+!> the eigen decomposition Hp = V W V^T of the projected, symmetrised Hessian.
+subroutine compute_compliance(unit, H, B, xyz, natoms, nint, C, stat)
+   !> Formatted output unit for the rank diagnostic.
+   integer, intent(in) :: unit
+   !> Number of atoms.
```
thfroitzheim, line 369: Here again

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   !> Cartesian Hessian in Hartree/Bohr^2, dimension (3*natoms, 3*natoms).
+   real(wp), intent(in) :: H(3*natoms, 3*natoms)
+   !> Wilson B matrix, dimension (nint, 3*natoms).
+   real(wp), intent(in) :: B(nint, 3*natoms)
+   !> Cartesian reference coordinates in Bohr, dimension (3, natoms).
+   real(wp), intent(in) :: xyz(3, natoms)
+   !> Compliance matrix, dimension (nint, nint), in Bohr^2/Hartree.
+   real(wp), intent(out) :: C(nint, nint)
+   !> Status: zero on success, the LAPACK info of the failing DSYEV otherwise.
+   integer, intent(out) :: stat
```
thfroitzheim, line 382: And we can use the error type for error propagation

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+
+   integer :: i, j, ndim, lwork, info, nrigid, nvib, rank_h
+   real(wp) :: tol_h, normq, center(3), scale
+   real(wp), parameter :: eps_svd = 1.0e-10_wp
```
thfroitzheim, line 386: Comment on the choice

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   ! modified Gram-Schmidt; the rotation about the molecular axis of a
+   ! linear molecule produces no displacement
+   nrigid = 0
+   do i = 1, 6
+      do j = 1, i - 1
+         normq = dot_product(Q(:, i), Q(:, j))
+         Q(:, i) = Q(:, i) - normq * Q(:, j)
+      end do
+      normq = norm2(Q(:, i))
+      if (normq < 1.0e-8_wp*scale) then
```
thfroitzheim, line 421: Lets at least make this a parameter and is this value general?

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   end do
+   nvib = ndim - nrigid
+
+   ! H averaged into symmetry first: the numerical Hessian handed over by
+   ! the frequency code is only symmetric to within its finite-difference
+   ! noise
+   Hp = 0.5_wp * (H + transpose(H))
+
+   ! Hp = (1 - Q Q^T) Hp (1 - Q Q^T)
+   call dgemm("N", "N", ndim, 6, ndim, 1.0_wp, Hp, ndim, Q, ndim, 0.0_wp, T1, ndim)
```
thfroitzheim, line 436: Lets use the blas wrappers in mctc

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   call dgemm("N", "T", ndim, ndim, 6, -1.0_wp, T1, ndim, Q, ndim, 1.0_wp, Hp, ndim)
+   call dgemm("T", "N", 6, ndim, ndim, 1.0_wp, Q, ndim, Hp, ndim, 0.0_wp, T2, 6)
+   call dgemm("N", "N", ndim, ndim, 6, -1.0_wp, Q, ndim, T2, 6, 1.0_wp, Hp, ndim)
+   Hp = 0.5_wp * (Hp + transpose(Hp))
+
+   ! Hp = V W V^T, overwriting Hp with V
+   lwork = -1; allocate(work(1))
+   call dsyev("V", "U", ndim, Hp, ndim, W, work, lwork, info)
```
thfroitzheim, line 444: Here the lapack wrapper

src/compliance.f90
```diff
@@ -0,0 +1,476 @@
...
+   ! C = B V D V^T B^T = Z (Z D)^T,  Z = B V,  D = diag(1/w_i)
+   ! Modes of the projected Hessian that are still zero are dropped.
+   tol_h = eps_svd * maxval(abs(W))
+   rank_h = count(abs(W) > tol_h)
+   if (rank_h /= nvib) then
+      write(unit, "(A,I0,A,I0)") &
+         & "  Note: Hessian rank=", rank_h, " /= 3N-rigid=", nvib
+   end if
+   call dgemm("N", "N", nint, ndim, ndim, 1.0_wp, B, nint, Hp, ndim, 0.0_wp, Z, nint)
+   ZD = Z
```
thfroitzheim, line 461: What is happening here?

src/internals/redundant.f90
lmseidler: yes, there is nothing speaking against this. specifically internals are used in geosum.f90 and intmodes.f90, i think they both have their own definitions and there is a lot of redundant code. but this PR is already pretty large and i wanted to do more refactors maybe step by step in order not to convolute too many things.

src/internals/graph.f90
lmseidler: according to a quick look over the naming conventions in xtb:

Across non-ignored Fortran files, counting derived-type declarations:
 - TPascalCase: 91
 - snake_case: 71
 - Neither: 13

So it's basically 50:50. since tblite uses snake_case, i would prefer snake case for new additions to the source code. i think the TPascalCase convention is inspired by dftb+

src/internals/redundant.f90
lmseidler: to keep the signature as simple as possible and to minimize dependencies (this method only needs the cartesian coords, no need to import also the molecule type) i think it's preferable to pass only the single attribute.

src/internals/type.f90
lmseidler: i think fortran doesn't have an enum type. this way is already the most simple way of implementing this. of course you could create a type with just these params as attributes, but i think this doesn't do much in terms of readability.

src/compliance.f90
lmseidler: i would here again prefer to pass only the two primitives instead of the whole type. in principle nat could also be left out but keeping it for readability.

src/compliance.f90
```diff
@@ -0,0 +1,541 @@
...
+      Q(:, i) = Q(:, i) / normq
+   end do
+   nvib = ndim - nrigid
+
+   ! H averaged into symmetry first: the numerical Hessian handed over by
+   ! the frequency code is only symmetric to within its finite-difference
+   ! noise
+   Hp = 0.5_wp * (H + transpose(H))
+
+   ! Hp = (1 - Q Q^T) Hp (1 - Q Q^T)
```
thfroitzheim, line 497: Is this actually our mctc_gemm interface? Don't we have anything shorter (also for Lapack?)

thfroitzheim (dismissed):

man/xtb.1.adoc
```diff
@@ -254,6 +254,20 @@ NOTE: You can only select *one* runtyp, only the first runtyp will be used
 *--o1nh*::
     perform the numerical hessian calculation using the ODLR approximation (O1NumHess)
 
+Hessian runs automatically perform a compliance analysis for nonperiodic
```
thfroitzheim: Sorry just saw it now. Please indent this part (and maybe reduce a bit this is quite long)

thfroitzheim (approved):

Issue #1445, opened 2026-09-11
Constraints Depend on Atom Order


hagenmuenkler:
**Describe the bug**
When using xtb to run a constrained optimization, using the "all torsions=true" option to automatically constrain all torsion angles, the result depends on the way the atoms are ordered in the input .xyz file. In some cases, I see no constraints applied at all. 

**To Reproduce**
I have attached two input files (github didn't allow me to upload .xyz or .inp files, so they come as .txt files in the upload)

[butane_heavy_first.txt](https://github.com/user-attachments/files/32104626/butane_heavy_first.txt)
[butane_reversed.txt](https://github.com/user-attachments/files/32104637/butane_reversed.txt)
[constrain.txt](https://github.com/user-attachments/files/32104801/constrain.txt)

The input files have 
a) butane with a typical atom order, in which the heavy atoms appear first
b) butane with a reversed order, in which all hydrogen atoms appear first
c) constrain.inp has the constraints with the "all torsions=true" block and a force constant 

Compare the output of two runs: 
xtb butane_heavy_first.xyz --input constrain.inp --opt --verbose
and 
xtb butane_reversed.xyz --input constrain.inp --opt --verbose 

The output for the first run shows no constrains being applied, whereas the output for the second run shows 12 constrained torsion angles. The STDOUT output of the two runs shows what constraints are applied

[xtb_heavy_first.log](https://github.com/user-attachments/files/32105487/xtb_heavy_first.log)
[xtb_reversed.log](https://github.com/user-attachments/files/32105488/xtb_reversed.log)

This can also be confirmed by looking at the output .xyz files (again as .txt)

[xtbopt_heavy_first.txt](https://github.com/user-attachments/files/32105593/xtbopt_heavy_first.txt)
[xtbopt_reversed.txt](https://github.com/user-attachments/files/32105594/xtbopt_reversed.txt)

**Expected behaviour**
Both runs should behave the same, and the behavior the runs show in the "reversed" atom ordering is the expected one. In this case, the deliberately distorted middle torsion angle of is maintained. 

**Additional context**
[...]

PR #1451, opened 2026-09-18
Fix for constrained optimization - Issue 1445

hagenmuenkler:
This PR provides:
-  a fix for the problem described in Issue 1445, which describes that the code to extract torsion angles currently depends on the order of atoms in the input
- a small change for the documentation of the constraints for the "elements" block. The old version could give the impression that elements would create positional constraints

I have not added tests for the changes to the $constrain logic. My own offline tests show that the changes fix the issue I had described.


hagenmuenkler:
Resolves #1445

hagenmuenkler:
The issue this was linked to was fixed by #1450, which applies the minimal fix to get_bonds. 

This PR aims to only have a single function "approx_bonds" (already used in other places) to infer bonds, but there are differences in behavior between the former "get_bonds" function and "approx_bonds". I'll wait for a decision for the maintainers whether these two functions should be reduced to 1, simplifying design and testing. If not, we can close the PR.

 Indeed, the current state of the PR suffers from circular import problems

PR #1452, opened 2026-09-19
CI: Add source-dist workflow

Growl1234:
Add a self-contained source distribution to the xtb CI.
- Build `xtb-*.tar.xz` with `meson dist --include-subprojects` for offline installation.
- Upload the source archive as a release artifact.
- Rename the existing bleed Linux binary archive to `xtb-bleed-linux-x86_64.tar.xz`.
- Reserve `xtb-bleed.tar.xz` for the Meson source distribution.


thfroitzheim (commented):
I see the point of renaming the linux version, but won't the source distribution just duplicate the source release already present in the bleeding-edge release (e.g., [here](https://github.com/grimme-lab/xtb/releases/tag/bleed))

Growl1234:
If you are referring to the "Source code", I don't think it's duplication. This `meson dist` tarball bundles its subprojects and makes users possible to process an offline installation from scratch, just like what tblite, dftd4, etc. provide.

thfroitzheim (approved):

Growl1234:
I'm wondering if `sha256.txt` is still needed? That file seems to be stale.

